PR #2, opened 2025-12-11
API deployment

raindrift:
Python API deployment scripts for GCP. mirrors the scripts in the golang repo.

This also includes the secrets for elasticsearch, even though we don't technically need them yet. We will very soon.

Finishes #96

### DIFF
--- a/scripts/apictl.sh
+++ b/scripts/apictl.sh
@@ -1,0 +1,239 @@
+#!/bin/bash
+
+# apictl - Simple control script for API services
+
+set -e
+
+# Configuration
+REGION="us-east1"
+SERVICES=(
+    "greenearth-api"
+)
+
+# Colors for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
+
+# Check if a service exists
+service_exists() {
+    local service=$1
+    gcloud run services describe "$service" --region="$REGION" > /dev/null 2>&1
+}
+
+# Get service status (simplified)
+get_service_status() {
+    local service=$1
+
+    if ! service_exists "$service"; then
+        echo "NOT_DEPLOYED"
+        return
+    fi
+
+    # Check manual instance count (0 = stopped)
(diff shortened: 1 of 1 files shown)


scripts/deploy.sh
```diff
@@ -0,0 +1,250 @@
...
+    # Resource and scaling configuration
+    deploy_cmd="$deploy_cmd --min-instances=$API_INSTANCES_MIN"
+    deploy_cmd="$deploy_cmd --max-instances=$API_INSTANCES_MAX"
+    deploy_cmd="$deploy_cmd --cpu=1"
+    deploy_cmd="$deploy_cmd --memory=512Mi"
+    deploy_cmd="$deploy_cmd --timeout=60"
+    deploy_cmd="$deploy_cmd --concurrency=80"
+
+    # Allow unauthenticated access (adjust based on your needs)
+    deploy_cmd="$deploy_cmd --allow-unauthenticated"
```
mjmor: Perhaps we should wait to deploy until we have a basic auth token or something simple like that? Or will the service URL that gets exposed via deployment somehow be internal-only?

scripts/gcp_setup.sh
```diff
@@ -0,0 +1,246 @@
...
+    # gcloud projects add-iam-policy-binding "$PROJECT_ID" \
+    #     --member="serviceAccount:$sa_email" \
+    #     --role="roles/run.invoker" \
+    #     --condition=None
+
+    # If you need to access GCS buckets, add this:
+    # gcloud projects add-iam-policy-binding "$PROJECT_ID" \
+    #     --member="serviceAccount:$sa_email" \
+    #     --role="roles/storage.objectViewer" \
+    #     --condition=None
```
mjmor: delete?

mjmor (approved):
Trust your judgement on this, but just curious: should the internal API be internal on the VPC instead of exposed on a public URL? Or are we going to do that later, and we're not concerned about exposing on public for now?

scripts/deploy.sh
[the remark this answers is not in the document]
raindrift: But in the meantime, I'll turn the service off since we also don't need to pay for it yet.

PR #53, opened 2026-04-13
Ranker (Engagement Prediction) Endpoint in API

### LINKED ISSUE #13: Build user features during inference
For the moment, we're going to build the user features from raw data on-the-fly. Eventually we'll persist the user features so they're fast and so that we can base them on more data than what we have stored. But for now let's just do the slow thing.

For this story to be done, we should be running the ES query to get whatever user data will be used to compute the user features, and then supplying it to the feature encoding code. We have later stories for generating candidates, and then running the model against those candidates.

davidfreifeld:
Closes #13

Added /rank/models and /rank/predict which mimic /candidates/. We now have two implemented rankers, a default/fallback one that just ranks the candidates based on the candidate generation score, and one that calls queries the inference server two tower model endpoints.

The two tower ranker takes in a user_did and the candidates exactly as output from the /candidates/generate/ endpoint and then queries the user tower endpoint with the history of embeddings of the user's recent likes, and the post tower endpoint with all of the candidate posts. Then it takes the dot product of the user embedding with each of the post embeddings, and returns the posts in decreasing order of that final score. This logic (which is arguably the most important thing to review here) is in predict() in src/app/lib/rankers/two_tower.py.

Additional changes: 
- I noticed there was ELASTICSEARCH_URL and GE_ELASTICSEARCH_URL, and the same with ELASTICSEARCH_API_KEY and GE_ELASTICSEARCH_API_KEY, so I consolidated them into just the GE_ versions. If we wanted these to stay separate I can revert the change.
- Similarly I consolidated CandidateGenerateResponse and CandidateGenerateResult

Notes:
- For now we have 2 new env vars (GE_INFERENCE_MAX_HISTORY_LEN and GE_INFERENCE_EMBED_DIM) specified in .env.example (and I set them in my local .env file for testing). They should match the corresponding values on the inference server. Though I imagine we want a better way to ensure that these are consistent. We should discuss, and could add the update to this story or a new one.
- If there is an error in the two tower rank computation, the API will currently return an error and no results. We could also have it just return the list of candidates, with or without the ranking based on the candidate generation score (and use the fallback ranker to do this).
- There is a DEFAULT_LIKED_POSTS_LIMIT in elasticsearch.py, which is set to 50. We might want to also line this up with the max history length configured for the inference server / model.


scripts/deploy.sh
```diff
@@ -15,7 +15,7 @@ REGION="us-east1"
 ENVIRONMENT="stage"
 
 # Elasticsearch configuration
-ELASTICSEARCH_URL="INTERNAL_LB_PLACEHOLDER"
```
raindrift, line 18: Thanks for taking care of this.  I keep thinking I got every instance of it, and then there's another one I missed.

src/app/lib/rankers/candidate_score_test.py
```diff
@@ -0,0 +1,55 @@
...
+@pytest.fixture
+def ranker():
+    return CandidateScoreRanker()
+
+
+def test_name(ranker):
+    assert ranker.name == "candidate_score"
+
+
+def test_predict_ranks_by_descending_score(ranker):
```
raindrift, line 20: nit: some candidate generators won't provide a score. The random one is an example that we have now, but there will probably be others. It looks like that case is handled, but it might be nice to have a test for this case too. Presumably all candidates with no score just go to the end, which is honestly fine. Like this is probably just a ranker we'll use for testing anyway.

src/app/lib/rankers/predict.py
```diff
@@ -0,0 +1,53 @@
...
+    if ranker is None:
+        raise RankModelNotFoundError(model_name)
+
+    if any(candidate.at_uri is None for candidate in request.candidates):
+        raise RankerError("All candidates must include at_uri")
+
+    if model_name == TWO_TOWER_MODEL and not request.user_did:
+        raise RankerError("user_did is required for two_tower")
+
+    deduped_candidates = dedup_candidates(request.candidates)
```
raindrift: Is this the right place for candidate deduplication? I would think that internally, the function producing the candidates should be responsible for that. Then again, this answers an api endpoint that can be called with arbitrary data.

But if the user wants to send us duplicated candidates, perhaps we should just rank the duplicates.

Part of why I bring this up is because there are some weird cases where having a guarantee that the number of inputs and outputs are the same is useful. For example, if we want to re-rank some set of inputs and log how the positions of the items changed. Maybe we don't need to care about that though.

Thoughts?

src/app/lib/rankers/two_tower.py
```diff
@@ -0,0 +1,243 @@
...
+        raise RankerExecutionError(
+            TWO_TOWER_MODEL_NAME,
+            "GE_INFERENCE_MAX_HISTORY_LEN environment variable is required",
+        )
+
+    embed_dim_raw = os.environ.get("GE_INFERENCE_EMBED_DIM")
+    if not embed_dim_raw:
+        raise RankerExecutionError(
+            TWO_TOWER_MODEL_NAME,
+            "GE_INFERENCE_EMBED_DIM environment variable is required",
```
raindrift: If we're specifying the model name in a constant in the code, it's probably ok to also put the embedding dimension in a constant, since that's a property of the model. Or do we expect it to change a lot, such that (eg) it'll be different on stage and prod? I ask because maintaining configuration in the environment is harder, so we should only do it if it's a secret or if it varies between deployments.

src/app/lib/rankers/two_tower.py
```diff
@@ -0,0 +1,243 @@
...
+    post_embeddings: list[list[float]],
+    *,
+    base_url: str,
+    api_key: str,
+) -> list[list[float]]:
+    url = f"{base_url}/models/post-tower/predict"
+    headers = {"X-API-Key": api_key}
+    payload = {"post_embeddings": post_embeddings}
+
+    async with httpx.AsyncClient(timeout=30.0) as client:
```
raindrift, line 70: It doesn't have to be in this PR, but I suspect that pretty soon we'll want to make a client for the inference service to make calls like this a little cleaner. If we want to retain some type checking, we can pull in the pydantic models that the inference service is using for its inputs.

But I'm mostly just mentioning it now so we can hopefully remember to do it later. Right now there are few enough call sites that I don't think it matters.

src/app/lib/rankers/two_tower.py
```diff
@@ -0,0 +1,243 @@
...
+        self, 
+        es,
+        user_did: str | None,
+        candidates: list[CandidatePost]
+    ) -> RankerResult:
+        inference_base_url, inference_api_key, inference_max_history_len, inference_embed_dim = (
+            get_inference_settings()
+        )
+        
+        ####### USER #######
```
raindrift, line 113: Something to consider (again, not for this PR, just mentioning it) is that we can call the user and candidate post sections of this pipeline in parallel, up until that last dot product. Let's keep it simple for now though, since this is our very first model and it'll be easier to make changes if we don't have to deal with concurrency bugs.

But if you agree that this might be a useful optimization, perhaps we should make a story for it so we remember to do it later.

src/app/routers/rank_error_test.py
```diff
@@ -0,0 +1,45 @@
+"""Tests for rank router error mapping."""
```
raindrift: nit: it's probably ok to put this stuff in rank_test.py.

raindrift (approved):
This looks good! I have a few minor comments but overall I think this is pretty much ready to go. Once it's merged let me know and I can follow up with a second PR to add a feed that uses the ranker. Let's also deploy it then so we can use the endpoint to test the connection between services.

src/app/lib/rankers/two_tower.py
davidfreifeld: Yeah the embedding dimension I don't expect to change very much at all (certainly not for a long time). So I'll hard-code it. 

The other similar env var though is GE_INFERENCE_MAX_HISTORY_LEN. Which is a property of the model, and I could see it changing across deployments as we experiment with the number of likes to use. Do you have thoughts on how we might keep that in sync between the two services? One idea would be to have an API endpoint on the inference server that returns the loaded user tower's max history length. Then this API would query it before running inference. But feels a little clunky/slow. Thoughts?

src/app/lib/rankers/two_tower.py
davidfreifeld: Cool - just made a story: https://github.com/greenearth-social/api/issues/54

src/app/lib/rankers/two_tower.py
davidfreifeld: Cool, good point. I made a story: https://github.com/greenearth-social/api/issues/55

In the story I wrote this: 
>Our current version of the ranker is the two tower model. Since we're not using it for candidate generation, we're not using it in the standard two tower-y way (where you cache post embeddings and just calc the user embedding) - we're just generating the user and all candidate post embeddings at the same time. If we keep using it like this, we should update the code to calculate these in parallel, as an optimization. If, before we need this optimization, we end up moving to a stage where we use the two tower for candidate gen and a heavier model for ranking, then we probably won't need this story.

src/app/lib/rankers/predict.py
davidfreifeld: Yeah I think you're right. The candidate generation phase also dedupes before returning. And we'll have a "diversity" pass after ranking where we could potentially confirm the absence of duplicates if we want to. So I'll remove it.

src/app/lib/rankers/candidate_score_test.py
davidfreifeld: https://github.com/greenearth-social/api/pull/53/changes/fbda28be36223a02a36cb1151a92acbc9e728423

src/app/lib/rankers/predict.py
davidfreifeld: https://github.com/greenearth-social/api/pull/53/changes/fbda28be36223a02a36cb1151a92acbc9e728423

src/app/routers/rank_error_test.py
davidfreifeld: https://github.com/greenearth-social/api/pull/53/changes/fbda28be36223a02a36cb1151a92acbc9e728423

src/app/models.py
```diff
@@ -98,6 +99,41 @@ class CandidateGenerateResult(BaseModel):
...
+    candidates: list[CandidatePost] = Field(
+        ...,
+        min_length=1,
+        description="Candidates to rank in the same shape returned by /candidates/generate",
+    )
+    model: str | None = Field(
+        None,
+        description="Optional ranking model identifier. Defaults to the service default.",
+    )
+    user_did: str | None = Field(
```
mjmor: I understand that not all models need this, but wouldn't it be straightforward to always require it anyway. I imagine the caller always has it, and always requiring simplifies the interface.

mjmor (approved):
@raindrift covered all the good stuff - I just had a minor nit (feel free to skip and merge if you're already ready to do so). This is amazing. Exciting!

src/app/models.py
davidfreifeld: Yeah - I was going back and forth on this. But I think you're right. Just updated it: https://github.com/greenearth-social/api/pull/53/changes/980bfc8f20edbdfff8137f6b35d71b64420e0a9d

src/app/lib/rankers/two_tower.py
raindrift: Interesting question. My first thought is that if possible we should make it robust to being out-of-sync, but in the case where that would significantly degrade performance we should instead make it fail loudly when it's wrong.

As far as where to store the configuration: if it's deployment-level config, the environment is probably the right place. Redeploying services to change config like this is fine. But if it's the kind of config we expect to change more often, we could also keep it someplace where all the services can see it. Firebase is probably the right place in that case. But there are tradeoffs with all of these.

raindrift: Oh but to follow up a little more on that: the main reason to put it in a database (like firebase) is that when a service starts up it can publish its capabilities, and then other services can reference that when calling it.

Having an endpoint for requesting those parameters isn't a terrible idea either, especially if the caller caches the result for a minute or two.

I suppose in all cases, we need to be able to tolerate it being wrong, or accept some failures sometimes. Since deployments are non-atomic, we can't guarantee that we will have a single version of the service running while a deployment is happening.

Regardless, I think we want to have a shared pattern for this kind of inter-service configuration. Like if we build something for this, let's make sure we the same thing everywhere.

davidfreifeld: Makes sense. 

But now I'm re-thinking the boundary of the two services though. Maybe it shouldn't be the caller's (i.e. the GreenEarth API's) responsibility to get the input in the right shape. Maybe that should happen inside the inference service. We could still have a maximum number of liked posts that we fetch for a user's history in the API. But then the inference service would be in charge of taking the variable-length history, and either truncating it to the maximum history length the model expects, or padding it with zeros up to that length. 

Now that I'm thinking about it, I feel like that separation actually makes a little more sense anyway. And it would remove the need for the GreenEarth API to know the inference service's/model's configured max history length. Wdyt? cc: @mjmor

mjmor: I agree with your last suggestion, Dave. History length sounds like it should be internal to the inference service / abstracted from the main API. If it were some sort of tunable option available to the main API server, that'd be a different story. 

That being said, I'd suggest we leave it as a environment variable for now and file a story to refactor later (e.g., right before we start experimenting with number of likes to use). That way we can merge and test e2e / MVP

davidfreifeld: Thanks Max. I like that plan. 

(No rush but) whenever one of you has a chance to take a look at my updates to deploy.sh and gcp_setup.sh, then I will merge and deploy, thanks!

scripts/deploy.sh
```diff
@@ -15,7 +15,11 @@ REGION="us-east1"
...
+GE_ELASTICSEARCH_URL="INTERNAL_LB_PLACEHOLDER"
+
+# Inference configuration
+GE_INFERENCE_BASE_URL=""
```
mjmor: Could we accept whatever is set in the environment, like so:

Someone calling the deploy script:
```
export GE_INFERENCE_BASE_URL="https://foo.bar.baz"
./scripts/deploy.sh
```

Line 21 here would be:
```
GE_INFERENCE_BASE_URL="${GE_INFERENCE_BASE_URL:-}"
```

That way we don't necessarily have to update this script, just the local / not committed `.env` file once we have the stable URI for the inference server working.

mjmor: Oh, now I see the empty default is overwritten with a commandline arg. Could we still move to the proposed above? Then priority order will be:
1. Command line arg
2. Existing environment var
3. Default empty string (two-tower fails)

mjmor (approved):
LGTM :) just one minor nit @davidfreifeld

scripts/deploy.sh
davidfreifeld: Makes sense to me - just updated: 355b69d1375d5c454fd15413651b0754f1998542

src/app/lib/rankers/two_tower.py
davidfreifeld: https://github.com/greenearth-social/engagement-prediction/issues/158

PR #62, opened 2026-04-20
Use stable inference service hostname

raindrift:
This PR uses the stable inference service hostname to call the inference service. It also includes a new script for testing the rank endpoint end-to-end to make sure the connection between services is functioning properly.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -41,6 +41,13 @@ export GE_FIRESTORE_PROJECT="demo-no-project"
 
 ########## Inference Variables ##########
 
-export GE_INFERENCE_BASE_URL="https://engagement-prediction-inference-url.run.app"
+# Optional explicit override. In stage/prod Cloud Run deploys, the API deploy script
+# now defaults to mapped domains when this is unset:
+#   stage -> https://inference-stage.greenearth.social
+#   prod  -> https://inference.greenearth.social
+# Keep this set in local development when you want to call a local inference service.
+export GE_INFERENCE_BASE_URL=""
+export GE_INFERENCE_DOMAIN="inference-stage.greenearth.social"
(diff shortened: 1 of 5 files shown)


.env.example
```diff
@@ -41,6 +41,13 @@ export GE_FIRESTORE_PROJECT="demo-no-project"
...
 
-export GE_INFERENCE_BASE_URL="https://engagement-prediction-inference-url.run.app"
+# Optional explicit override. In stage/prod Cloud Run deploys, the API deploy script
+# now defaults to mapped domains when this is unset:
+#   stage -> https://inference-stage.greenearth.social
+#   prod  -> https://inference.greenearth.social
+# Keep this set in local development when you want to call a local inference service.
+export GE_INFERENCE_BASE_URL=""
+export GE_INFERENCE_DOMAIN="inference-stage.greenearth.social"
+export GE_ENABLE_INFERENCE_DOMAIN_MAPPING=true
```
mjmor, line 51: This multi-layered fallback logic is sort of hard to follow. Could we simplify and just require that the inference domain is set in the environment or as a command line flag whenever deploy.sh is called?

mjmor (approved):
Looks good. Just a minor comment on simplifying a bit.

.env.example
raindrift: I'm not a huge fan of specifying it every time, since it means that you have to think about this whenever you run a deployment. Since it changes infrequently, I'd like for it to be a thing that lives in the repo. However, I will see if I can make the code simpler.

raindrift: Ok, I removed most of the fallback code from deploy.sh and just made it automatic. There's really very little reason to ever change it.

mjmor: Perfect that sounds great thanks for doing that

PR #69, opened 2026-04-28
Remove shared/input_data_helpers call and shared dependency

davidfreifeld:
## Broader Change Background
This PR is part of a broader change affecting 3 repos: engagement-prediction, api, and inference-service. The main architectural change is to implement [the proposal in the second section of this diagram](https://docs.google.com/drawings/d/1i2rEPGKf3jSY46oTOsRJXfWu9xNiXwq43COJrID0K5U/edit?usp=sharing) - in short, move the inference service code to its own repo. 

The inference service *should* live in its own repo, as the engagement-prediction repo is mainly for model training. Eventually we might like to move the `shared` package into its own repo as per the above diagram. The initial motivation for this change was to decouple the model and input data logic from the Green Earth API, and move it into the inference service, where it belongs. This allows the API to no longer depend on the `shared` package and to no longer need to keep track of configs like the maximum user history length (which we also need to keep track of in the inference service). Initially I tried moving the `shared` input data helpers call to inference_service within the engagement-prediction repo, but ran into a constraint having to do with the requirements of docker and cloud run directory structure. And hence extracted inference service into its own repo. Across these 3 PRs we will:
- Create a new `inference-service` repo
- Move the shared dependency and the input data padding and masking from `api` -> `inference-service`
  - The new `inference-service` repo will use pipenv with a Pipfile and Pipfile.lock to manage the env 
- Clean up the handling of raw user history inputs:
  - Explicitly allow an empty list [], a list with one empty list [[]], a single (non-batched) list of lists [ [float, float, ...], [float, float, ...], ...], or a batched history, a list of list of lists: [ [ [float, float, ...], [float, float, ...], ...], ...]. In the last case, we allow empty histories, e.g. [ [], [ [float, float, ...], [float, float, ...], ...], ...]
  - This logic is validated in a function in the `shared` package, and made explicit in the `inference-service` app

## This PR
- Removes the call to the input data helper function in the api app. This is now done inside of inference service.
- Therefore we can remove the dependency on the `shared` package altogether, as well as the env var GE_INFERENCE_MAX_HISTORY_LEN

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -50,4 +50,3 @@ export GE_INFERENCE_BASE_URL=""
 export GE_INFERENCE_DOMAIN="inference-stage.greenearth.social"
 export GE_ENABLE_INFERENCE_DOMAIN_MAPPING=true
 export GE_INFERENCE_API_KEY="inference-service-api-key-here"
-export GE_INFERENCE_MAX_HISTORY_LEN=128
--- a/Pipfile
+++ b/Pipfile
@@ -11,7 +11,6 @@ elasticsearch = "==9.3.0"
 aiohttp = "*"
 atproto = "*"
 google-cloud-firestore = "*"
-shared = {git = "https://github.com/greenearth-social/engagement-prediction.git", ref = "main", editable = true}
 
 [dev-packages]
 pytest = "*"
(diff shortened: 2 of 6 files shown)


mjmor (approved):
Much cleaner

PR #73, opened 2026-05-04
Followed users candidate generation

davidfreifeld:
Closes #74

## This PR
Created a new candidate generator called followed_users. It hits the bsky API endpoint to get the accounts that the requesting user follows. Then it queries ES to get the most recent N posts from those users. That's it - there's no more fancy logic for ranking, scoring, diversifying etc.

The bsky getFollows API endpoint has a max limit of 100. The code as written hits the endpoint with a limit of 100, to retrieve up to 1,000 users. For each request, if it fails, it can retry up to 1 time. If the querying fails altogether midway through, we just take the users we've retrieved so far, and use those to query for posts. All of this logic is a good starting point, I think, but I am open to feedback if we'd like to change any of it.

As for the feeds, I added "followed_users" as a generator in the "basic-similarity" feed, with a weight of 0.4. This of course can be changed if we want to do something different.

As described in the story, this PR does not attempt to cache followed users on a given user. In my experience with playing around with the feed, it seems that the similarity candidate generator might be slower than the followed users generator. But this is anecdotal and not scientific. Perhaps we can deploy this version when it's ready and play around with it some more, and then determine if we need to add caching? Open to advice here. 

### Other small changes
- Pulled the name of the MiniLM sentence transformer into a variable, so it's not hard-coded in a number of places. This added a number of touched files to the PR, but they all have minimal changes.
- There was some shared processing code of the retrieved candidates at the end of each candidate generator's "search" function, so I abstracted that away into a reusable function.
- Removed the bit in the README for this repo about the `shared` dependency, since it is no longer relevant - it was removed in a prior PR.

## Testing
Ran the API locally and successfully generated candidates for my own DID and another random one. Also published the new basic similarity feed and verified that it contains a mix of similar posts and posts from accounts I follow.


src/app/lib/candidates/followed_users.py
```diff
@@ -0,0 +1,234 @@
...
+# ---------------------------------------------------------------------------
+
+# Maximum number of followed users to use in the query
+MAX_FOLLOWED_USERS = 1_000
+
+# Maximum page size accepted by app.bsky.graph.getFollows (should not be changed)
+FOLLOWS_PAGE_LIMIT = 100
+
+# Maximum total time spent paginating followed users for one request
+FOLLOWS_LOOKUP_TIMEOUT_SECONDS = 5.0
```
mjmor: I think we'll want to reduce this to one second at most. If blue sky is taking longer than that, we might consider caching as we had originally discussed. Either way, this probably isn't going to scale well for users with many, many followers. Maybe we should just make a follow-up issue for now. Open to your thoughts.

src/app/lib/candidates/followed_users.py
```diff
@@ -0,0 +1,234 @@
...
+
+    raise AssertionError("unreachable follow lookup retry state")
+
+
+async def get_followed_user_dids(user_did: str, limit: int) -> list[str]:
+    base_url = "https://public.api.bsky.app/xrpc/app.bsky.graph.getFollows"
+    followed_dids: list[str] = []
+    cursor: str | None = None
+
+    if limit <= 0:
```
mjmor, line 80: Do you know if the follow list is sorted and stable? Or if we have a limit and the user requests the feed multiple times, is it possible they get a different subset of followed users in their feed each time?

mjmor (approved):
Looks good. Just left wondering what the performance is. What have you seen in testing? If we're seeing something that's less performant than needed, even for our accounts with not many followers, we'll probably want to rethink how we're doing caching or time limits or followed account post limits.

Approving now because I think it's probably better to just merge it in and let everyone test the feeds on their own to get a sense of performance. We can debug later and tweak things as necessary during dogfooding.

Also, thanks for all the refactors and additional tests. Very helpful.

src/app/lib/candidates/followed_users.py
davidfreifeld: I can't find any documentation about the sorting or stability of the list. The results don't returned a "followedAt" timestamp or anything like that, so can't say for certain if it's sorting by that.

Anecdotally, the sorting seems stable from query to query.

src/app/lib/candidates/followed_users.py
davidfreifeld: Yeah, I think we probably want a follow-up story for caching. I'll make one now. 

Mark Hamill () has ~2,500 follows (who knew Luke loved Bsky so much?). With no time limits or max users in place, it takes about 2-2.5 seconds for the API to return candidates for him. 

Another user (did:plc:f6odp2tkcw4tiojcqbukjbr2) has 240,000 follows(!). Without the limits it takes 2 minutes 15 seconds to get results. 

For my did (only 49 follows), the results get returned very quickly.

raindrift (approved):
This looks good! I think we want a separate story to maintain a follower cache, since the limit makes this kind of inefficient. I'll write it up now.

src/app/lib/candidates/followed_users.py
```diff
@@ -0,0 +1,234 @@
...
+                or not _is_retryable_follow_lookup_error(exc)
+            ):
+                raise
+            await asyncio.sleep(FOLLOWS_RETRY_BACKOFF_SECONDS)
+
+    raise AssertionError("unreachable follow lookup retry state")
+
+
+async def get_followed_user_dids(user_did: str, limit: int) -> list[str]:
+    base_url = "https://public.api.bsky.app/xrpc/app.bsky.graph.getFollows"
```
raindrift, line 76: Oh! You know, if all we need are the user DIDs, we might be able to get better results from the `com.atproto.repo.listRecords?repo=<did>&collection=app.bsky.graph.follow` endpoint instead. The getFollows endpoint returns hydrated user records, which I expect to be kind of slow, and might explain the 100 record limit. But `listRecords` should pull the follows directly from the user's repo, which in theory is much faster.

We don't get much data about those users though, aside from their DID and the date they were followed. But maybe that's enough?

davidfreifeld: I actually explored that endpoint - and surprisingly I found that it was slower!

raindrift: wat. ok that's weird. perhaps they already have a cache in front of the more detailed one

davidfreifeld: Here is what I saw:

getFollows summary
==================
Runs: 2
Follow counts: [2566, 2566]
Page counts: [27, 27]
Mean elapsed: 0.965s
Min/Max elapsed: 0.312s / 1.617s

listRecords summary
===================
Runs: 2
Follow counts: [2656, 2656]
Page counts: [28, 28]
Mean elapsed: 4.109s
Min/Max elapsed: 3.973s / 4.245s

raindrift: Wow damn, that is *way* slower.

davidfreifeld: Yeah and actually that's not even including the fact that for the listRecords method you have to find the PDS endpoint for the DID first, which takes additional time.

PR #75, opened 2026-05-04
Switch KNN searches to `posts_recent` index and add timing telemetry

### LINKED ISSUE #59: Switch over API to posts_recent index for KNN + add telemetry
## Issue

#315 reduced KNN search latency by at least 10x by creating a separate index for the most recent posts (last 1 week of posts) in `posts_recent` index. 

Our feed API code is still querying `posts` index alias which queries all post data and is extremely slow for KNN, effectively breaking the production GreenEarth Feed that relies on KNN.

There's also a possibility that the new recent posts KNN query still is too slow for production, but we can't be certain without telemetry.

## Solution

Update the API implementation code to query the `posts_recent` index alias instead of `posts` index alias wherever we call KNN (candidate generator and skylight endpoint, also grep for more). Consider whether we should consolidate these under a KNN helper method if we call KNN more than a couple times so people don't accidentally mess this up in the future.

Add telemetry around the KNN query in the API implementation code. We should also create a basic timer API call handler (e.g, python context manager for the API call) that can be reused to time similar API calls in the future.

raindrift:
The production feed was effectively broken for users relying on KNN-based ranking because all KNN queries were hitting the full `posts` index alias, which is extremely slow for nearest-neighbour search. This switches both KNN call sites to the `posts_recent` alias (last ~1 week of posts), which was created in greenearth-social/ingex#315 specifically for this purpose.

Note: embedding lookups by URI (to build the query vector from a user's liked posts) intentionally remain on the full `posts` index, since liked posts may be older than a week.

A `POSTS_KNN_INDEX` constant is introduced in `lib/elasticsearch.py` as the single source of truth for the KNN index name, so future callers can't accidentally use the wrong index.

A reusable `timed` async context manager is added in `lib/telemetry.py` and applied to both KNN call sites, so we can observe actual latency in production logs and verify the improvement.

Closes #59

### DIFF
--- a/src/app/lib/candidates/post_similarity.py
+++ b/src/app/lib/candidates/post_similarity.py
@@ -5,15 +5,16 @@ Generates candidates by finding posts similar to a user's recent likes:
 1. Query the ``likes`` index for the user's most recent liked posts.
 2. Fetch those posts from the ``posts`` index to retrieve MiniLM L12 embeddings.
 3. Average the embeddings into a single query vector.
-4. Run a kNN nearest-neighbours search against the ``posts`` index.
+4. Run a kNN nearest-neighbours search against the ``posts_recent`` index.
 """
 
 import logging
 
 from ...models import CandidatePost
 from .base import CandidateGenerator, CandidateResult
-from ..elasticsearch import unwrap_es_response, fetch_recent_liked_post_uris, fetch_post_embeddings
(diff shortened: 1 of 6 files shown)


mjmor (approved):
yay!

mjmor:
Did it fix the production similarity feed?

PR #81, opened 2026-05-07
Diversification pass on generated feeds

max-ai-ast:
Closes #80

This PR implements a Maximal Marginal Relevance (MMR) diversification pass on the feed pipeline to reduce same-author and topic-similar clustering. The pass runs after candidate generation and score-based sorting, as the final step before results are returned.

# This PR

- Adds `author_did: str | None` field to `CandidatePost` and populates it from Elasticsearch `_source.author_did` in all three candidate generators (`post_similarity`, `popularity`, `random_posts`)
- Implements greedy MMR in `src/app/lib/diversify.py`: selects candidates maximizing `(1 - BETA) * relevance - BETA * max_similarity_to_selected`, where similarity combines author identity and (when `AUTHOR_WEIGHT < 1.0`) embedding cosine distance. 
  - Sets AUTHOR_WEIGHT to 0.75 so same author posts are likely removed unless their content is maximally different. we might want to play with this though, because sometimes you want multiple related posts from an author in feed like when someone is doing a tweet chain on a topic and circumventing the max character limit.
  - Also, note: @davidfreifeld : I used "the maximum similarity between an item and any item from the set" from your comment, but we might consider coming back and using the smoothing option as you described, since will achieve max diversity which may not be optimal.
- Wires `mmr_rerank` into both the fresh-generation and pagination-regeneration branches of `getFeedSkeleton` in `xrpc.py`
- Exposes a standalone `POST /diversify` endpoint for external use

# Testing

- `pipenv run pytest` — 231 tests pass
- `src/app/lib/diversify_test.py`: empty input, single candidate, same-author spread (bob precedes alice/2), all-different-author relevance order
- `src/app/routers/diversify_test.py`: auth required, empty 200, all candidates preserved, same-author spread via HTTP
- `src/app/routers/xrpc_test.py`: added `test_same_author_candidates_are_spread_in_feed` verifying feed output interleaves same-author posts

### DIFF
--- a/src/app/lib/diversify.py
+++ b/src/app/lib/diversify.py
@@ -8,53 +8,88 @@ from ..models import CandidatePost
 BETA = 0.5
 AUTHOR_WEIGHT = 0.75
 
+
 def mmr_rerank(candidates: list[CandidatePost]) -> list[CandidatePost]:
     if len(candidates) <= 1:
         return list(candidates)
 
+    n = len(candidates)
     max_score = max((c.score or 0.0) for c in candidates)
-    if max_score > 0.0:
-        norm_scores = [(c.score or 0.0) / max_score for c in candidates]
-    else:
-        norm_scores = [0.0] * len(candidates)
+    norm_scores = (
+        [(c.score or 0.0) / max_score for c in candidates]
+        if max_score > 0.0
+        else [0.0] * n
+    )
 
+    # Pre-decode embeddings once so the inner loop never repeats base64 work.
+    vecs: list[list[float] | None] = []
+    for c in candidates:
+        if c.minilm_l12_embedding is not None:
+            try:
+                vecs.append(decode_float32_b64(c.minilm_l12_embedding))
+            except Exception:
+                vecs.append(None)
+        else:
+            vecs.append(None)
+
+    author_dids = [c.author_did for c in candidates]
+    remaining = list(range(n))
     selected: list[int] = []
-    remaining: list[int] = list(range(len(candidates)))
+    # max_sim[i] tracks the highest similarity candidate i has to any selected
+    # candidate so far. Updated incrementally — one new comparison per remaining
+    # item each round instead of recomputing the full max from scratch.
+    max_sim = [0.0] * n
 
     while remaining:
         if not selected:
             best = max(remaining, key=lambda i: norm_scores[i])
         else:
-            def mmr_score(i: int) -> float:
-                rel = (1 - BETA) * norm_scores[i]
-                sim = BETA * max(_similarity(candidates[i], candidates[j]) for j in selected)
-                return rel - sim
-            best = max(remaining, key=mmr_score)
+            best = max(
+                remaining,
+                key=lambda i: (1 - BETA) * norm_scores[i] - BETA * max_sim[i],
+            )
 
         selected.append(best)
         remaining.remove(best)
 
+        for i in remaining:
+            s = _raw_similarity(author_dids[i], author_dids[best], vecs[i], vecs[best])
+            if s > max_sim[i]:
+                max_sim[i] = s
+
     return [candidates[i] for i in selected]
 
 
-def _similarity(a: CandidatePost, b: CandidatePost) -> float:
-    if a.author_did is not None and a.author_did == b.author_did:
-        author_match = 1.0
-    else:
-        author_match = 0.0
-
-    if (
-        AUTHOR_WEIGHT < 1.0
-        and a.minilm_l12_embedding is not None
-        and b.minilm_l12_embedding is not None
-    ):
-        vec_a = decode_float32_b64(a.minilm_l12_embedding)
-        vec_b = decode_float32_b64(b.minilm_l12_embedding)
+def _raw_similarity(
+    author_a: str | None,
+    author_b: str | None,
+    vec_a: list[float] | None,
+    vec_b: list[float] | None,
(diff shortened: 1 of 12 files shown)


davidfreifeld:
The diversify endpoint here differs from the rank and candidates endpoints in that it's just currently set up to only support a single version: MMR. Whereas we have multiple candidate generators and rankers. I imagine that we might experiment with other types of diversification in the future, so it might be nice to set this endpoint up to support that? But I think it's also reasonable to set that up later when we add a second diversification method - I don't feel too strongly.

mjmor:
@davidfreifeld , I think you're saying this endpoint could accept a diversification algorithm type, of which MMR is one (and initially the only) type? 

I also don't feel strongly about it and would lean towards omitting that for now for the sake of merging and getting something working end-to-end. If we decide we add multiple diversification algorithms in the future, the algorithm_type API parameter can be nullable, where the null default is treated as MMR to support any existing callers without breaking (if we even have callers at that point).

mjmor:
Deployed on stage and working quite well. Confirmed no same author posts back to back and threaded posts continue to render fine. Content seems diverse and unit tests should confirm the same.

davidfreifeld:
Yup, that's exactly what I was saying. 

I'm fine with leaving it as-is for now to get the pipeline working end-to-end 👍 

I'll make a story for supporting multiple diversification endpoints just so we don't forget about it.

src/app/lib/diversify.py
```diff
@@ -0,0 +1,101 @@
...
+
+def mmr_rerank(candidates: list[CandidatePost]) -> list[CandidatePost]:
+    if len(candidates) <= 1:
+        return list(candidates)
+
+    n = len(candidates)
+    max_score = max((c.score or 0.0) for c in candidates)
+    norm_scores = (
+        [(c.score or 0.0) / max_score for c in candidates]
+        if max_score > 0.0
```
davidfreifeld: I believe it's possible for all the scores to be negative, in which case the max would be negative as well, and we'd erroneously use all zeros instead of the actual scores. So maybe we could instead do min-max scaling here? 

normalized = (score - min_score) / (max_score - min_score)

src/app/lib/diversify.py
```diff
@@ -0,0 +1,101 @@
...
+        else:
+            vecs.append(None)
+
+    author_dids = [c.author_did for c in candidates]
+    remaining = list(range(n))
+    selected: list[int] = []
+    # max_sim[i] tracks the highest similarity candidate i has to any selected
+    # candidate so far. Updated incrementally — one new comparison per remaining
+    # item each round instead of recomputing the full max from scratch.
+    max_sim = [0.0] * n
```
davidfreifeld: We initialize the max_sim's at 0 but then only re-write them below if the raw similarity is greater than 0. But cosine similarity can be negative. It wouldn't be *wrong* per se to do it this way - it just means we're treating all negatively-similar posts the same. If we want to instead boost the most anti-similar posts, we could initialize the max_sim's at negative infinity for example.

src/app/routers/xrpc.py
```diff
@@ -325,7 +328,9 @@ async def get_feed_skeleton(
         gen_request, request.app.state.es, swallow_errors=True
     )
 
-    all_uris = [c.at_uri for c in result.candidates if c.at_uri]
+    candidates = sorted(result.candidates, key=lambda c: c.score or 0.0, reverse=True)
+    candidates = mmr_rerank(candidates)
+    all_uris = [c.at_uri for c in candidates if c.at_uri]
```
davidfreifeld, line 333: Here (and in the regeneration path) - we might want to filter out no-URI candidates first? So that a no-URI candidate doesn't tarnish the MMR results unnecessarily?

src/app/routers/xrpc.py
```diff
@@ -325,7 +328,9 @@ async def get_feed_skeleton(
         gen_request, request.app.state.es, swallow_errors=True
     )
 
-    all_uris = [c.at_uri for c in result.candidates if c.at_uri]
+    candidates = sorted(result.candidates, key=lambda c: c.score or 0.0, reverse=True)
```
davidfreifeld, line 331: Here (and in the regeneration path) - do we need to sort before passing to mmr_rerank?

src/app/lib/diversify.py
mjmor: Good catch thank you!

src/app/routers/xrpc.py
mjmor: I think it's a safe invariant that all candidates will have an at_uri.

src/app/routers/xrpc.py
mjmor: I believe so. Since MMR is greedy, the order prior to running MMR has an impact on the result order.

davidfreifeld: (This isn't a big deal so we don't need to go too deep on it but) I guess the only way that I see it mattering is breaking ties. For two items with the same weighted author/similarity score, the algo will take the one that had the initially higher relevance score. Otherwise I don't think the initial order has an effect. Because (aside from ties) it always selects the remaining item with the highest weighted score, regardless of order. So if the initial sorting takes any meaningful amount of time, I'm not sure we need it.

mjmor: It's linear against a constant set size so I'll leave it in place.

davidfreifeld (approved):

PR #84, opened 2026-05-13
Two tower feed

raindrift:
## Two-tower ranked feed

Wires the two-tower inference model into a new `ranked` feed that uses `post_similarity` for candidate generation and the two-tower model for ranking.

### What changed

**New `ranked` feed** — `post_similarity` candidates (with `popularity` infill) are scored by the two-tower model before MMR diversification. The ranking pipeline is extracted into `_run_ranking_pipeline`, which handles the generate → rank → diversify sequence and is used for both fresh generation and cache-miss paths.

**`FeedConfig` extended** with `rank_request_template` (optional) and `diversify` (default `True`). When `rank_request_template` is set, candidates are passed through the named model before being returned. When `diversify=False`, MMR is skipped (used by the `random` feed).

**Batch splitting for post-tower calls** — the inference service has `GE_INFERENCE_MAX_BATCH=32` but the ranker was sending all candidates in a single request, causing 422s. Candidate embeddings are now sent in chunks of 32. Error responses now log the response body to make future inference errors easier to diagnose.

**Avoid redundant ES fetch for candidate embeddings** — candidate posts already carry `minilm_l12_embedding` from the generator step (both KNN hits and popularity results include it). The ranker now decodes it in memory rather than re-querying ES, falling back to a fetch only for any candidates missing it.

**`model_construct` for feed templates** — feed configs no longer need
placeholder values for session-specific fields (`user_did`, `candidates`).

### DIFF
--- a/src/app/lib/rankers/two_tower.py
+++ b/src/app/lib/rankers/two_tower.py
@@ -13,6 +13,7 @@ import httpx
 from ...models import RankedCandidate, CandidatePost, RankPredictResult
 from .base import Ranker, RankerExecutionError, RankerResult
 from ..elasticsearch import fetch_post_embeddings, fetch_recent_liked_post_uris
+from ..embeddings import decode_float32_b64
 
 
 logger = logging.getLogger(__name__)
@@ -138,9 +139,24 @@ class TwoTowerRanker(Ranker):
         ####### CANDIATE POSTS #######
         valid_candidates = [candidate for candidate in candidates if candidate.at_uri is not None]
         candidates_by_uri = {candidate.at_uri: candidate for candidate in candidates if candidate.at_uri is not None}
-        
-        # Get the embeddings for all the posts
-        candidate_embedding_pairs = await fetch_post_embeddings(es, list(candidates_by_uri))
+
+        # Use embeddings already carried on CandidatePost when available (avoids an ES round-trip).
+        candidate_embedding_pairs: list[tuple[str, list[float]]] = []
+        missing_uris: list[str] = []
+        for uri, candidate in candidates_by_uri.items():
+            if candidate.minilm_l12_embedding:
+                try:
+                    vec = decode_float32_b64(candidate.minilm_l12_embedding)
+                    candidate_embedding_pairs.append((uri, vec))
+                    continue
+                except Exception:
+                    pass
+            missing_uris.append(uri)
+
+        if missing_uris:
+            fetched = await fetch_post_embeddings(es, missing_uris)
+            candidate_embedding_pairs.extend(fetched)
+
         if not candidate_embedding_pairs:
             logger.info(
                 "No embeddings found for %d candidate posts of user %s",
--- a/src/app/lib/rankers/two_tower.py
+++ b/src/app/lib/rankers/two_tower.py
@@ -37,6 +37,9 @@ def get_inference_settings() -> tuple[str, str]:
     return base_url, api_key
 
 
+POST_TOWER_BATCH_SIZE = 32
+
+
 async def predict_post_tower_batch(
     post_embeddings: list[list[float]],
     *,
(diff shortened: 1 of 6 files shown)


src/app/lib/rankers/two_tower.py
```diff
@@ -125,9 +139,24 @@ async def predict(
...
-        # Get the embeddings for all the posts
-        candidate_embedding_pairs = await fetch_post_embeddings(es, list(candidates_by_uri))
+
+        # Use embeddings already carried on CandidatePost when available (avoids an ES round-trip).
+        candidate_embedding_pairs: list[tuple[str, list[float]]] = []
+        missing_uris: list[str] = []
+        for uri, candidate in candidates_by_uri.items():
+            if candidate.minilm_l12_embedding:
+                try:
+                    vec = decode_float32_b64(candidate.minilm_l12_embedding)
```
mjmor, line 149: We cache the decoded embeddings in the diversification pass because this step was very inefficient (albeit the diversification pass was exponential, not linear time complexity), so if we do see performance issues in prod, using shared decided embedding caches across steps would be a good place to start.

src/app/models.py
```diff
@@ -143,8 +142,20 @@ class FeedConfig(BaseModel):
...
     """
 
     display_name: str = Field(..., max_length=12)
     description: str = ""
     gen_request_template: CandidateGenerateRequest
+    rank_request_template: RankPredictRequest | None = Field(
+        None,
+        description="When set, candidates are ranked by this model before being returned.",
+    )
+    diversify: bool = Field(True, description="When False, MMR reranking is skipped.")
```
mjmor, line 161: We might not always use MMR, so we could say something more generic here if we wanted to future proof.

mjmor (approved):

src/app/lib/rankers/two_tower.py
```diff
@@ -37,6 +38,9 @@ def get_inference_settings() -> tuple[str, str]:
     return base_url, api_key
 
 
+POST_TOWER_BATCH_SIZE = 32
```
davidfreifeld, line 41: Good catch to start batching here. Just pointing out that I chose the max batch size of 32 in the inference service somewhat arbitrarily. So this is something we can also tune if prod perf isn't great.

src/app/routers/xrpc.py
```diff
@@ -113,6 +114,39 @@ class FeedSkeletonResponse(BaseModel):
...
+    if not candidates:
+        return []
+
+    if feed_cfg.rank_request_template is not None:
+        rank_req = feed_cfg.rank_request_template.model_copy(
+            update={"candidates": candidates, "user_did": gen_request.user_did}
+        )
+        rank_result = await run_predict(rank_req, es)
+        # Reorder CandidatePosts by model rank so MMR diversifies starting
+        # from the highest-ranked items.
```
davidfreifeld, line 140: nit: FWIW, I don't think the initial ranking actually matters for MMR. So you could possibly only sort in the non-diversify case. But it's probably not a huge speedup. Max and I were discussing this a bit in the PR for diversify too.

davidfreifeld: Actually - there's a bigger issue here I think. mmr_rerank() uses candidate.score. Whereas the ranking step saves the score in .rank_score, which is what we'd want to use. So there are a couple ways to fix this, but one option is to modify mmr_rerank so that it accepts the scores to use explicitly.

The point above still stands though - the ordering of the list that you pass into MMR shouldn't have an effect (aside from breaking ties I guess).

src/app/routers/xrpc.py
```diff
@@ -113,6 +114,39 @@ class FeedSkeletonResponse(BaseModel):
...
+    candidates = result.candidates
+
+    if not candidates:
+        return []
+
+    if feed_cfg.rank_request_template is not None:
+        rank_req = feed_cfg.rank_request_template.model_copy(
+            update={"candidates": candidates, "user_did": gen_request.user_did}
+        )
+        rank_result = await run_predict(rank_req, es)
```
davidfreifeld, line 138: run_predict will fail completely if any of the candidates don't have a uri. We might want to actually change that behavior or instead filter out candidates with missing uri's first?

PR #86, opened 2026-05-14
Return ranker results for users with empty like histories

### LINKED ISSUE #85: Always return results for rank/two_tower
Right now, if we don't have any like history for the given user, the inference service's two tower ranker just returns an empty list. We should probably have it return the candidates, with the score equal to the candidate generation score if it exists

davidfreifeld:
Closes #85 

If a user has no like history, the ranker currently returns an empty set. So a user with no like history wouldn't get a feed at all. 

The model already specifically has handling and training for users with no like history. So we just need to pass in an empty like history to the inference service, instead of short-circuiting and returning no results. 

I verified this on stage with my DID where I have no likes, and got results. Also added regression tests.

### DIFF
--- a/src/app/lib/rankers/two_tower.py
+++ b/src/app/lib/rankers/two_tower.py
@@ -91,24 +91,23 @@ class TwoTowerRanker(Ranker):
         
         ####### USER #######
         # 1. Get recently liked post URIs
+        user_history_vectors = []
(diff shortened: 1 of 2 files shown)


mjmor (approved):
Simple enough 🫡 thanks for the fast fix

PR #88, opened 2026-05-14
Return results when candidates are missing their content embeddings

### LINKED ISSUE #87: Return ranker results for candidates with missing embeddings
If candidates are missing content embeddings in elasticsearch, still return them, but with rank_score=None. Instead of dropping them silently.

davidfreifeld:
Closes #87 

If candidates are missing content embeddings in elasticsearch, still return them, but with rank_score=None. Instead of dropping them silently.


mjmor (approved):
Nice catch and thanks for adding the test 🙏

PR #95, opened 2026-05-19
Do not select replies as candidates

raindrift:
### Reply filtering

All four candidate generators (`post_similarity`, `popularity`, `followed_users`,
`random_posts`) now exclude reply posts by adding
`{"exists": {"field": "thread_parent_post"}}` to each query's `must_not` clause.
The posts index stores `thread_parent_post` on any post that is a reply, so
this reliably excludes them without requiring a separate boolean field.


raindrift:
This doesn't really change much, so I'm going to go ahead and merge it so I can get our obvious bugs fixed on prod.

davidfreifeld:
Looks good to me 👍

mjmor (approved):

PR #117, opened 2026-05-21
Network Likes Candidate Generator

### LINKED ISSUE #104: Candidate generator: network likes (posts liked by followed users)
We have a candidate generator for posts written by followed users, but we should have one for posts liked by followed users too.

If we have the ability when implementing this, also include posts from 2nd degree connections (users my follows follow). This part will probably depend on #30 or some partial implementation thereof.

Also consider the cold-start case.

davidfreifeld:
Closes #104 

## This PR
Adds a new candidate generator that (1) queries bsky for users that the target users follows, (2) queries ES for recent likes by those users, and (3) again queries ES to get those posts. It is pretty straightforward, with the exception being that from step 2 to step 3, we could lose posts for a variety of reasons: the post is not in our posts index, or the post is a reply on a thread, for example. So we can't just query `num_candidates` likes - we'll end up with too-few posts at the end. So this PR queries likes in pages until it has `num_candidates` final posts. 

The ordering (and candidate.score) of the posts is based on the number of likes. The post with the most likes from your followed users will be shown first, etc. The tie-breaker is based on which post had a more recent last like.

Note: this candidate generator is not yet used in any feeds, and there is no standalone feed for it yet either. 

Other small changes: 
- Moved the bluesky API querying functions (for getting followed users) from the followed_users feed file to its own file - bsky.py - since we're re-using it now.
- Changed fetch_recent_liked_post_uris() to be able to accept a list of did's instead of just one.
- Updated post_similarity.py to use the existing candidate_posts_from_es_response() function instead of reproducing the logic in it.
- Added a linter ignore comment in feeds.py because the whole file had erroneous red squigglies for me.

## Testing
- Added CI tests: bsky_test.py, network_likes_test.py
- Ran the API locally and generated candidates for myself


raindrift (approved):
This seems good! I'm a little concerned with some of the query redundancy here, but there's no way around it without changing things about how our db is structured. Which we may want to do, but let's do some profiling first and make sure it's necessary.

mjmor (approved):
@raindrift another thing we could try before refactoring the database is an API-side cache of posts and followers. Won't solve cold start but I bet a lot of the data fetched in-network doesn't change often. 

Not something for this PR, just continuing the dialogue on what we might do post-profiling.

PR #118, opened 2026-05-22
Remove prod feed prefix and shorten stage prefix to GE

mjmor:
Feed display names on Bluesky are limited to 24 characters. The previous `GreenEarth` production prefix consumed 11 characters (leaving only 13 for the feed name), and it was inconsistent with the `GE`-prefixed dev/stage conventions.

# This PR

- Removes the `GreenEarth` prefix for production environment feeds (prod feeds now display their base name only)
- Shortens the stage prefix from `GE Stg` to `GE`
- Adds `TestBlueskyDisplayNameLimit` test that asserts every feed × every environment stays within Bluesky's 24-character display name limit
- Updates the inline comment in `feeds.py` to reflect the new prefix rules

# Testing

- `pipenv run pytest scripts/publish_feed_test.py` — all 39 tests pass
- New `test_all_feeds_fit_bluesky_limit_in_all_environments` covers all 4 feeds × 3 environments (12 combinations)


src/app/feeds.py
```diff
@@ -13,11 +13,11 @@
 
 from .models import CandidateGenerateRequest, FeedConfig, GeneratorSpec, RankPredictRequest
 
-# NOTE: display_name is limited to 24 chars, including the prefix ("GreenEarth, GE Dev, or GE Stg")
+# NOTE: display_name is limited to 24 chars, including the prefix ("GE Dev" or "GE"; prod has no prefix)
 FEEDS: dict[str, FeedConfig] = {
     "basic-similarity": FeedConfig(
         display_name="Similarity",
```
raindrift, line 19: I think we need to give these better names if we're not going to have the GreenEarth prefix.

raindrift (commented):
Honestly, I'm not sure it's a good idea to do only this story without reworking our feed naming a bit. I started a Discord thread for it though, rather than discuss it here on github.

mjmor:
Will open a fresh PR for this.

PR #119, opened 2026-05-26
Feed generation optimization, first pass

raindrift:
The `ranked` feed's `getFeedSkeleton` is taking up to ~30s per request. A read of the call path surfaced several places where the handler was paying the sum of independent latencies (instead of the max), re-fetching data it already had, opening a fresh TLS connection per inference call, and gating the response on analytics writes. None of these are individually a 30s-class problem — but collectively they account for several seconds of wall clock on every request.

This PR is four self-contained commits, each addressing one of those.

## Commits

**`run independent candidate generators and inference calls concurrently`**
- `run_generate` now dispatches all generators via `asyncio.gather` so e.g. `post_similarity` and `followed_users` (which makes an outbound call to `public.api.bsky.app`) overlap instead of running back-to-back.
- `TwoTowerRanker.predict` runs the user-side fetch + user-tower inference concurrently with the candidate-side fetch + post-tower inference.
- `predict_post_tower_batch` fires its 32-embedding chunks in parallel on the shared client instead of sequentially.

**`add per-request cache for ES query results`**
- New `ContextVar`-backed `RequestCache`. `_run_ranking_pipeline` enters a `request_cache_scope()`, and the existing ES helpers (`fetch_recent_liked_post_uris`, `fetch_post_embeddings`) consult it transparently — no signature changes anywhere else.
- This removes the duplicate ES round-trips where `post_similarity` and the two-tower ranker were both fetching the user's recent likes and their embeddings. The cache is single-flight (per-key `asyncio.Lock`) so concurrent identical lookups collapse into one query.
- Future components that query the same data automatically benefit; no plumbing required.

**`share a single httpx.AsyncClient across requests`**
- Previously the two-tower ranker and the Bluesky follows lookup each constructed a fresh `httpx.AsyncClient` per invocation, paying a TLS handshake every time. A single ranked-feed request opened 3+ TLS connections (2 to inference, 1 to `public.api.bsky.app`).
- Hoist a process-wide `AsyncClient` managed by the lifespan handler. The bsky path now passes its tight per-request timeout via the per-call kwarg.

**`move session-recording Firestore writes off the request path`**
- The handler was awaiting, in series before the pipeline started: a DID document fetch (for username) + a Firestore read+write to upsert the user + a Firestore read+write to upsert feed activity. Pure analytics overhead gating every response.
- Spawn the resolve + upserts as a background task that overlaps with the pipeline.

## Behavior change (intentional)

Firestore session-write failures used to return `500 "Firestore write failed"`. They now log an error and return the normal `200` response, since these are analytics writes and aren't needed to compute the feed. The three tests that asserted the old contract were updated to assert the new logged-but-non-fatal contract. If you'd rather keep the fatal behavior, the simplest fix is to await the background task at the end of the handler.

### DIFF
--- a/src/app/lib/request_cache.py
+++ b/src/app/lib/request_cache.py
@@ -1,0 +1,64 @@
+"""Per-request cache for expensive lookups.
+
+A ``ContextVar`` holds the current request's cache so helper functions
+(e.g. the Elasticsearch query wrappers in ``lib/elasticsearch.py``) can
+consult it transparently without callers passing a cache parameter
+through every layer.
+
+The router enters ``request_cache_scope()`` at the top of each request;
+the scope is per-task, so concurrent requests get independent caches
+and child tasks spawned via ``asyncio.gather`` inherit the parent's
+cache automatically.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import contextlib
+from contextvars import ContextVar
+from typing import Any, Awaitable, Callable
+
+_current_cache: ContextVar["RequestCache | None"] = ContextVar(
+    "ge_request_cache", default=None
+)
+
+
+class RequestCache:
+    """Async-safe single-flight cache keyed by any hashable value.
+
+    Two concurrent calls with the same key share one underlying
+    computation, so duplicate ES queries within a request collapse
+    into a single round-trip.
+    """
+
+    def __init__(self) -> None:
+        self._store: dict[Any, Any] = {}
+        self._locks: dict[Any, asyncio.Lock] = {}
+
+    async def get_or_compute(
+        self, key: Any, factory: Callable[[], Awaitable[Any]]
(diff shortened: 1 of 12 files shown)


raindrift:
Note: this does appear to speed things up, but not by all that much. I suspect it's one of the underlying queries that is slow. But I wanted to handle this stuff before profiling it since we want all this stuff anyway, and I thought maybe it would be enough (it is not)

src/app/lib/request_cache.py
mjmor, line 1: fancccyyyy :) good idea

mjmor (approved):
Looks good - thanks for cleaning!

davidfreifeld:
Admittedly not my area of expertise but I looked through it all and it makes sense. All seem like very good improvements!

PR #121, opened 2026-05-28
Request profiling for bsky feed serving

raindrift:
## Motivation

The earlier perf pass on `raindrift/quick-performance-fixes` removed obvious code-read inefficiencies but didn't shrink the ranked feed's `getFeedSkeleton` latency enough. To know what to fix next, we need to attribute time to specific stages — ideally on real Bluesky AppView traffic that hits a locally-tunneled dev server, since synthetic curl calls don't reproduce the full request mix.

This PR adds two complementary profiling layers, both off by default in prod:

- **Stage-level `timed()` log lines** — semantic breakdown ("`run_generate` 800 ms, `run_predict` 2.3 s, `bsky_get_follows` 1.1 s") in the uvicorn console, enabled via `GE_TIMING_LOGS=1`.
- **`pyinstrument` flame chart per request** — written as HTML to `api/profiles/` when `GE_PROFILE=1` is set, for drilling into library-internal time that stage labels can't anticipate.

Both layers correlate via a server-generated 12-hex request ID so a single `grep` ties a waterfall to its flame chart file.

## Commits

**`add server-generated request ID and gate timed() logs on env var`**
- A Starlette middleware mints a request ID into a ContextVar and echoes it back as `x-request-id`. Inbound `x-request-id` headers are deliberately ignored to avoid log forging / correlation poisoning from untrusted callers.
- `timed()` now reads the ContextVar and emits `rid=<id>` when set, and **only emits the log line at all when `GE_TIMING_LOGS=1`**. The wrapped block always runs; production stays quiet by default.
- The inference HTTP calls forward the rid via the existing outbound headers dict, so once the inference service starts logging it we get cross-service correlation. The Bluesky public-API call is intentionally not modified — we don't send internal IDs to third-party services.

**`add timed() spans across the ranked feed path`**
- Spans at `run_generate`, `run_predict`, per generator, both two-tower side coroutines, both inference HTTP calls, the cached ES helpers, `popularity` / `followed_users` / `bsky_get_follows` calls, and the three feed-cache operations.
- Pure-CPU work (MMR, dedup, dot products, base64 decode) is intentionally not timed — those have always been microseconds.
- The two existing `timed()` callers (kNN searches) inherit the rid + env-var gate for free.

**`add pyinstrument middleware gated on GE_PROFILE=1`**
- Per-request flame chart at `interval=0.001`, `async_mode="enabled"` so `await` boundaries inside `asyncio.gather` are followed correctly.
- HTML files named `{ts}-{rid}-{path-slug}.html` in `api/profiles/` (gitignored). The directory and the filename embed the rid so a `grep rid=<id>` finds both the stage log lines and the matching profile file.
- pyinstrument is added to `[dev-packages]` with a lazy import inside `install_profiling`, so prod containers without the dev extras still start.
- Registered *before* `request_id_mw` so that Starlette's last-registered-first dispatch leaves the profiler nested inside the request-ID setup. That way the profiler sees the rid still set when building the output filename.

**`configure root logger so app INFO logs reach the console`**
- Uvicorn only attaches handlers to its own loggers, so anything emitted via `logging.getLogger("app.xxx")` — including the existing kNN spans, the new `timed()` lines, and `profile_written` — was being silently dropped. A `logging.basicConfig(..., force=True)` at the top of `main.py` fixes it. Level is overridable via `GE_LOG_LEVEL`.

## Behavior

| `GE_TIMING_LOGS` | `GE_PROFILE` | What you see |
|---|---|---|
| unset | unset | App startup logs + uvicorn access lines. `x-request-id` echoed on response. |
| `1` | unset | + per-stage waterfall in console, all sharing one `rid` per request. |
| unset | `1` | + an HTML flame chart in `api/profiles/` per request, plus a `profile_written rid=… path=…` log line. |
| `1` | `1` | Both. |

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -50,3 +50,12 @@ export GE_INFERENCE_BASE_URL=""
 export GE_INFERENCE_DOMAIN="inference-stage.greenearth.social"
 export GE_ENABLE_INFERENCE_DOMAIN_MAPPING=true
 export GE_INFERENCE_API_KEY="inference-service-api-key-here"
+
+########## Debugging Variables ##########
+
+# Set GE_TIMING_LOGS to "1" to enable logging of timing data for various operations in the API.
+export GE_TIMING_LOGS=""
+
+# Set GE_PROFILE to "1" to enable profiling of the API. Results will be stored in the
+# "profiles" directory.
+export GE_PROFILE=""
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ firebase-debug.log
 
 # pyinstrument flame charts written when GE_PROFILE=1
 profiles/
+
+# macOS metadata files
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,6 @@ requirements.txt
 firestore-debug.log
 firebase-debug.log
 .firebase/
+
+# pyinstrument flame charts written when GE_PROFILE=1
+profiles/
--- a/Pipfile
+++ b/Pipfile
@@ -17,6 +17,7 @@ pytest = "*"
 pytest-asyncio = "*"
 httpx = "*"
 ruff = "*"
+pyinstrument = "*"
 
 [requires]
 python_version = "3.13"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "d16b47f1e60a7f6e8979b6cc60d3e388551db5d10397a647c3051bd2e21cfddc"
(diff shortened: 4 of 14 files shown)


inthree3 (approved):
Oh, the extension of the `timed` method is exactly what I was doing. I didn't know that we can link its output with pyinstrument output via the 12-hex id and that sounds great!

The other ideas I hadn't thought of, including excluding pure-CPU functions, are cool too. Since my changes are a subset of yours, I'll pull this.

mjmor (approved):
All looks good but could you also add an entry in `.env.example` for the new debugging environment variables?

PR #127, opened 2026-05-28
Additional performance tuning.

raindrift:
> **Stacked on `raindrift/profiling`** — review & merge that PR first.

## Motivation

With the profiling instrumentation from the parent PR in place, the slow-query log and per-stage `timed()` waterfall made it possible to attack the ranked feed's remaining latency in a targeted way instead of guessing. Three real bottlenecks fell out:

1. **The kNN reply filter was triggering ES brute-force scoring.** With `must_not exists thread_parent_post` inside `knn.filter` (or in a sibling `bool.must_not`), the per-shard `vector_operations_count` jumped from a few thousand to ~118k, and ES ended up advancing through ~755k docs per shard checking for the field. p50 query time was ~860 ms; p90 was 2.6 s.
2. **The default `_source` was enormous.** A slow_es_query log line showed ES reporting `took=33 ms` on a query whose wall-clock elapsed was 2940 ms — almost 3 seconds spent shipping bytes. Posts carry ~13 source fields including TWO 384-dim embedding arrays (L12 + L6), so the default response was 746 KB for 50 hits and 2.3 MB for a 150-hit kNN.
3. **Embeddings were being shipped speculatively.** Most candidates that arrive from the generators never make it through dedup or aren't used by MMR/the ranker, so paying ~4-5 KB per candidate up front for embedding bytes is wasteful.

This PR addresses all three.

## Commits

**`prefilter posts_recent kNN search inside the knn clause`**
First attempt at #1 — moved the reply filter inside `knn.filter` to make ES pre-filter during HNSW traversal. Direct benchmarking against the cluster showed it didn't help; both pre- and post-filter forms triggered the brute-force fallback. Kept as a stepping stone for the next commit.

**`filter replies in Python rather than via the ES kNN filter`**
The actual fix. The kNN clause carries no reply-exclusion at all; we over-fetch 5x (capped between 60 and 500) and drop replies in Python after the kNN response. Empirically the kNN neighborhood of an averaged-likes vector is ~75% replies (much higher than the index baseline), so the over-fetch is necessary to still produce `num_candidates` results. p50 ES latency drops from ~860 ms to ~57 ms; p90 drops from ~2.6 s to ~1.1 s on this cluster. `exclude_uris` stays as an ES filter — it's a small `terms` clause against a keyword field, bitmap-friendly, and doesn't trigger the fallback.

**`log slow ES queries with the full request body for replay`**
A `SlowQueryLoggingES` wrapper around `AsyncElasticsearch.search()`. When wall-clock exceeds `GE_SLOW_ES_THRESHOLD_MS` (default 500 ms), the wrapper logs one WARNING with `rid`, `index`, `elapsed_ms`, and the full request body as JSON. Grep for `slow_es_query`, copy the `body=` payload into curl/Kibana, replay the query verbatim. This is what surfaced #2 and #3 above on real Bluesky AppView traffic.

**`restrict candidate ES queries to the _source fields we actually use`**
`CANDIDATE_SOURCE_FIELDS` is the minimum a candidate generator needs (`at_uri`, `author_did`, `content`, `thread_parent_post`, `contains_video`, and the L12 embedding). Applied as `_source=` on every `es.search()` in post_similarity, popularity, followed_users, random_posts, and network_likes. Wire bytes drop 2-5x; the win scales with the RTT/bandwidth between the API and the cluster.

**`defer embedding fetch until after candidate dedup`**
Removes the embedding from `CANDIDATE_SOURCE_FIELDS` entirely and adds a hydration pass in `_run_ranking_pipeline` that batches one `fetch_post_embeddings` call for just the candidates that survived dedup. A kNN at k=250 drops from ~1.86 MB to ~50 KB; the per-request cache lets the two-tower ranker piggyback on the hydration's results. Hydration failures are swallowed and logged — MMR falls back to author-only similarity and the two-tower ranker has its own refetch path, so a hiccup in the batch fetch doesn't fail the request.

## Tunables introduced

- `GE_SLOW_ES_THRESHOLD_MS` (default 500) — ms threshold above which a `search()` call gets logged with its full body. Re-read on each call so you can adjust it without restarting.
- `OVERFETCH_MULTIPLIER`, `MIN_OVERFETCH`, `MAX_OVERFETCH` in `post_similarity.py` — how aggressively to over-fetch kNN results to compensate for Python-side filtering. Defaults (5×, 60, 500) are tuned to the ~75% reply rate seen in averaged-likes neighborhoods on this cluster.


raindrift:
FYI, deploying this to stage to make sure it doesn't break stuff in a real environment (it works fine locally)

src/app/lib/candidates/post_similarity.py
```diff
@@ -46,36 +56,63 @@ async def knn_search_posts(
 ) -> list[CandidatePost]:
     """Run a kNN search against the ``posts_recent`` index and return candidate posts.
 
-    Uses the MiniLM L12 embedding field for nearest-neighbour matching. Each
-    hit is converted to a :class:`CandidatePost` with the ES score attached.
+    Reply exclusion (``thread_parent_post exists``) and the rare
+    ``video_only`` filter are applied **in Python** rather than via the ES
+    ``knn.filter`` parameter. Empirically, putting those filters in the
+    kNN clause forces ES into brute-force scoring (>1 s per shard, several
```
mjmor, line 62: I'm very surprised there's no way to do this type of pre-filyering in ES kNN without triggering brute force. I also trust your detective work, so just noting this for now (guess that's what we get for using ES as a one size fits all solution).

src/app/lib/candidates/post_similarity.py
```diff
@@ -46,36 +56,63 @@ async def knn_search_posts(
...
+    seconds total), even when ~55% of docs survive the filter. Profiling
+    shows the per-shard ``vector_operations_count`` jumps from a few
+    thousand to >100k when this happens.
+
+    Cheap, bitmap-friendly filters (``exclude_uris`` as a ``terms``
```
mjmor, line 67: Have we tried using `thread_parent_post` as a terms filter instead of inside kNN subquery? I believe it is indexed by default so should be supported as a terms filter.

mjmor (approved):
Looks good, just one more idea to try using terms filter on the thread parent post since it seems like that's the only way to trigger true prefiltering prior to kNN in Elastic Search. If it works, we could skip all the python-side filtering and avoid #359 (which I'm hoping we don't have to do). 


Perhaps you already tried this though. Either way - carry on if this is working! We can always try the terms filter as a pre-req to #359.

src/app/lib/candidates/post_similarity.py
raindrift: I am pretty sure I tried these things, but I'll make sure just in case.

inthree3 (approved):
Impressive on finding the problem's reason and solving approach! I've been reading through the thread all day but still feel the need to read more.

src/app/lib/candidates/post_similarity.py
raindrift: Ok, I looked into this more, and a terms filter won't work well because we don't have any kind of low-cardinality field describing whether something is a reply (like a boolean), nor do we have the thread_parent_post field on documents that do not have a parent. Like the field is omitted instead of set to "". So there's nothing to match on. We could fix this, but if we're messing with ingex anyway we might as well split the collection.

mjmor: Ah - that makes total sense. Thanks for looking into it!

PR #128, opened 2026-05-29
feat: support multiple API keys via Firestore

### LINKED ISSUE #103: Support multiple API keys
We're going to have multiple people using our public APIs, so we need to have a key for each distinct caller for good operational hygiene. 

Implement simplest method of provisioning / manging multiple API keys. This might even be a google secret manager dictionary (if supported) of allowed keys and some metadata about each key. Otherwise, we can actually put these in a database with some scripts to manage keys.

inthree3:
Closes #103

Implements the multi-key system discussed in the issue — 
each caller gets their own key so we can track and revoke 
access individually.

Keys are stored in Firestore (`api_keys` collection) as 
salted SHA-256 hashes. The plaintext key is shown once at 
generation and never stored anywhere.

What changed:
- `security.py` now looks up keys from Firestore instead 
  of comparing against a single `API_KEY` env var
- `scripts/apikeys.py` CLI to generate, list, revoke, 
  and check usage per key
- Each key stores the owner's email and tracks 
  monthly call count (auto-resets each month)
- Removed `API_KEY` secret from `deploy.sh`

## How to deploy

**Deployment order (important)**

Firestore keys must exist before the new code is live — 
deploying first will cause all API calls to return 401.

**1. Generate a key against the target Firestore DB**
```bash
GE_FIRESTORE_PROJECT=greenearth-471522 GE_FIRESTORE_DATABASE=greenearth-stage \
  pipenv run python scripts/apikeys.py generate 
```

**2. Deploy**
```bash
./scripts/deploy.sh --environment stage
```

**3. Verify with the generated key**
```bash
# Valid key → 200
curl https:/// -H "X-API-Key: gea_"
# Wrong key → 401
curl https:/// -H "X-API-Key: gea_wrongkey"
```

## How to test

**Unit tests**

```bash
python3 -m pytest -v  # 330 passed
```

**End-to-end with Firestore emulator**

Start the emulator (requires Firebase CLI and Java):

```bash
firebase emulators:start --only firestore
```

In a separate terminal, generate a key and inspect it:

```bash
GE_FIRESTORE_EMULATOR_HOST=127.0.0.1:8080 pipenv run python scripts/apikeys.py generate test@example.com
GE_FIRESTORE_EMULATOR_HOST=127.0.0.1:8080 pipenv run python scripts/apikeys.py list
GE_FIRESTORE_EMULATOR_HOST=127.0.0.1:8080 pipenv run python scripts/apikeys.py usage <key_id>
```

Start the dev server pointing at the emulator:

```bash
GE_FIRESTORE_EMULATOR_HOST=127.0.0.1:8080 GE_ELASTICSEARCH_API_KEY=dummy \
  pipenv run uvicorn src.app.main:app --reload
```

Confirm authentication works as expected:

```bash
# Valid key → 200
curl http://localhost:8000/ -H "X-API-Key: gea_<generated-key>"

# Wrong key → 401
curl http://localhost:8000/ -H "X-API-Key: gea_wrongkey"
```

Revoke the key and confirm it no longer works:

```bash
GE_FIRESTORE_EMULATOR_HOST=127.0.0.1:8080 pipenv run python scripts/apikeys.py revoke <key_id>
curl http://localhost:8000/ -H "X-API-Key: gea_<generated-key>"  # → 401
```


src/app/documents.py
```diff
@@ -51,3 +51,21 @@ class FeedActivityDocument(BaseModel):
...
+class ApiKeyDocument(BaseModel):
+    """An issued API key stored in the ``api_keys`` collection.
+
+    The document ID in Firestore is ``key_id``.
+    The plaintext key is never stored — only the salted SHA-256 hash.
+    """
+
+    key_id: str = Field(..., description="8 hex chars; also the Firestore document ID")
+    key_hash: str = Field(..., description="SHA-256(bytes.fromhex(salt) + full_key.encode()) as hex")
+    salt: str = Field(..., description="32 hex chars, random per key")
```
raindrift: In retrospect, I don't think a salt is necessary. I'm honestly not quite sure why I wrote it into the story. It does not meaningfully improve security, and it adds a bunch of extra complexity. Would it be a lot of trouble to remove it and just store the SHA256 hash of the key instead?

raindrift (commented):
This looks great, except for one issue where I think I didn't specify the task quite right.

src/app/documents.py
inthree3: Fair. 

The 48-character hex key is already strong enough against rainbow table attacks, so the salt is probably overkill. Thanks for pointing it out and I'll update it accordingly!

raindrift: Totally. And from a threat modeling perspective, the set of scenarios where an attacker has readonly (but not write) access to that collection is pretty small. If you can write to firebase, you can just make yourself an api key.

If we are at all concerned about rainbow table attacks though, the right thing would be to use bcrypt instead of SHA256. It's up to you whether it feels like it's worth the time to switch. I could honestly go either way.

raindrift (approved):
This is ready to merge, up to you if you want to switch to bcrypt or leave it as-is.

raindrift (changes requested):
Oh wait! Actually, can you document how to generate API keys in the README? And can that also make clear how to switch environments (so, if you want to make a key in dev, stage, or prod, how do you do that?)

mjmor (commented):
Looks good to me to but +1 for adding docs please :)

src/app/documents.py
inthree3: Good point on the threat model! 

One thing to consider on the bcrypt side, since we're hashing on every request, the added latency could be a concern as traffic grows. I'd lean toward keeping SHA-256 for now.

mjmor (approved):
image

raindrift (approved):
👍

PR #129, opened 2026-05-29
pass author dids to the two tower model call

davidfreifeld:
Closes #50 

## This PR
This update makes it so that the two tower ranker will always attempt to look up the author did of each target post and user liked history post, and pass it to the inference service (both the post tower and user tower respectively). If the author did is missing from the ES query, it passes an empty string, which will map to the "unknown" author index inside of the inference service. Note I added `fetch_post_embeddings_and_authors()` to elastcisearch.py instead of trying to extend the functionality of `fetch_post_embeddings()`. So the tradeoff is that the latter stays focused, but there is now a little bit of duplicated code, since both functions fetch the embeddings.

## Testing
Added tests and ran pytest. Ran API locally with updated inference-service successfully. Manually compared calcs for a few post embeddings, both with and without author did's, to calling the model directly.

### DIFF
--- a/src/app/lib/candidates/post_similarity_test.py
+++ b/src/app/lib/candidates/post_similarity_test.py
@@ -9,7 +9,8 @@ from ..candidates.post_similarity import (
     fetch_recent_liked_post_uris,
     knn_search_posts,
 )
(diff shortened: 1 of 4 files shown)


mjmor (approved):
Easy money

image

PR #130, opened 2026-05-29
sendInteractions endpoint support

### LINKED ISSUE #106: Support the sendInteractions endpoint to collect views, clicks, etc
Docs here (but they are not good): https://docs.bsky.app/docs/api/app-bsky-feed-send-interactions

To use this, we set `acceptsInteractions: true` on the feed record, then we create a `/xrpc/app.bsky.feed.sendInteractions` XRPC endpoint. It'll get some json, in the form described in the docs.

AFAICT, the user id is not sent with this data. However, we can supply a `feedContext` field when generating the feed, and it'll be sent back verbatim in the `sendInteractions` payload. Max length 2000.

So, feedContext should contain, at minimum, the user ID, along with any other feed metadata we want to include (a specific feed generation instance id, stuff we might want to log, whatever). This data should at least be signed/verified, or just encrypted using a secret. Otherwise a malicious actor could send invalid interactions to the otherwise public endpoint and poison the data.

Interactions we get (log all of them): requestLess, requestMore, clickthroughItem, clickthroughAuthor, clickthroughEmbed, interactionSeen, interactionLike, interactionRepost, interactionReply, interactionQuote, and interactionShare

I am thinking that we will want individual records in firestore for each interaction, rather than a single document containing them all. We may eventually move this data into ES, but firestore is fine for now.

Do we want to expire these eventually? No, not for now, but make sure we have a nice timestamp field so we can.

FYI, the docs for this end point are very bad and I asked an LLM instead

raindrift:
Closes #106.

## Motivation

Bluesky's AppView can forward user interaction signals (likes, reposts, replies, "show more/less", clickthroughs, seen events, …) to a feed generator that declares `acceptsInteractions: true` and implements `app.bsky.feed.sendInteractions`. These signals are valuable training data for engagement prediction. The catch: the payload carries no user DID — the only way to attribute an interaction is via a `feedContext` string we set when serving the skeleton, which the AppView echoes back verbatim. Because the endpoint is otherwise public, that string must be signed so a malicious caller can't forge interactions and poison the data.

## What's in this PR

- **`POST /xrpc/app.bsky.feed.sendInteractions`** — verifies each interaction's `feedContext`, drops forged/missing ones, and persists valid ones as individual auto-ID documents in a top-level `interactions` Firestore collection with `created_at` (anchor for any future TTL). Records are written in a background task; the endpoint returns `{}` immediately per the lexicon.
- **Signed `feedContext` tokens** (HMAC-SHA256, stdlib only, no new runtime deps). Each token carries `did`, `feed`, `rid`, `iat` and is attached to every item in a `getFeedSkeleton` response. The request id doubles as the feed-cache key, so served item order is recoverable from the existing cache within its TTL.
- **`GE_FEED_CONTEXT_SECRET`** — required at startup (`lifespan` raises if unset, matching the existing `GE_ELASTICSEARCH_API_KEY` check). Documented in `.env.example`.
- **`acceptsInteractions: true`** is now set on the published feed records (both the `publish_feed` and `sync_feeds` code paths — the sync path was previously building its own record dict without it).
- **Event names are stored without the `app.bsky.feed.defs#` lexicon prefix** (e.g. `interactionLike`, `requestMore`). Falls back to the original value when stripping would leave the empty string.

## Side updates bundled in

- **Per-environment log level default**: stage/prod → `WARNING` (suppresses INFO chatter from `timed()` spans, ES transport, httpx, etc. while keeping warnings, errors, and uvicorn access logs); local/dev → `INFO`. `GE_LOG_LEVEL` still overrides. Dropped the dead `LOG_LEVEL=info` env var from `deploy.sh`.
- **`pyright` added as a dev dependency** so we can run it locally. A `[tool.pyright]` config, CI step, and the handful of pre-existing test type errors will come in a follow-up PR.
- **`CandidatePost` field defaults** switched from `Field(None, ...)` to `Field(default=None, ...)` so pyright recognizes them. Behavior-neutral; cleared ~40 pre-existing pyright errors in unrelated test files.


src/app/routers/xrpc.py
```diff
@@ -195,6 +257,31 @@ def _get_feed_cache(request: Request) -> FeedCache:
...
+# ---------------------------------------------------------------------------
+# feedContext helpers
+# ---------------------------------------------------------------------------
+
+
+def _make_feed_context(user_did: str, feed_name: str, request_id: str) -> str:
+    """Build the signed feedContext token shared by every item in a response.
+
+    ``request_id`` doubles as the feed-cache key, so the served item order can be
+    recovered from the cache during its TTL window.
```
mjmor, line 269: Nice. (In the future) we can also store a sample of generated feeds for debugging allowing us to better debug candidate generators vs. rankers vs. diversification etc. and cross-reference with interaction data.

mjmor (approved):
YEAAAAAAAAA!!!! 

image

src/app/routers/xrpc.py
raindrift: We do store the generated feeds! At least, for a while. They go in the `feed_cache` collection in firebase. We need them to properly support cursoring.

PR #131, opened 2026-05-30
Custom GCP Monitoring metrics for feed render latency

### LINKED ISSUE #89: Keep track of how long it takes feeds to render
The time between receiving a feed request and returning a response is going to be one of our most important metrics. Let's make sure we have it on our dashboard, and that it's properly segmented by feed.

Cloud Run's built-in metrics won't cut it — we'll need custom instrumentation.

What Cloud Run gives us by default:

`run.googleapis.com/request_latencies` tracks wall-clock request time, but the only available dimension labels are `response_code`, `revision_name`, and `service_name`. There's no URL path or query parameter label. Even if there were a path label, it wouldn't help here — the feed type lives in a query parameter (`?feed=at://…/app.bsky.feed.generator/<feed_name>`), not the path. Every feed type hits the same path: `/xrpc/app.bsky.feed.getFeedSkeleton`

Also: include wallclock logging for the candidate generators because the performance will very greatly depending on user (at least for candidate generators like followed users generator).

mjmor:
Closes #89

Feed render latency is our most important performance metric but Cloud Run's built-in `run.googleapis.com/request_latencies` can't be segmented by feed type — the feed lives in a query parameter, not the URL path. This PR adds custom OTel metrics exported to GCP Cloud Monitoring.

# This PR

- Add `opentelemetry-api`, `opentelemetry-sdk`, and `opentelemetry-exporter-gcp-monitoring` dependencies
- New `MetricCollector` class (`src/app/lib/metrics.py`) — wraps OTel SDK; exports to GCP when `PROJECT_ID` is set, falls back to stdout for local dev; lazily creates instruments with type inference (`_count` → counter, `_rate` → gauge, else histogram)
- Extend `timed()` with `record_metric: bool = False` opt-in and `metric_attrs` for low-cardinality metric labels; `**extra` remains log-only so continuous values (counts, offsets) don't produce metric label explosions; `label` doubles as the metric name
- Wire `MetricCollector` into app lifespan (`main.py`), reading `PROJECT_ID`, `REGION`, and `ENVIRONMENT` to match existing env var conventions; shutdown on exit
- Instrument `xrpc.get_feed_skeleton` → `feed.render.duration_ms` histogram labeled by `feed_name`
- Instrument `generate._run_one` → `candidates.generate.duration_ms` histogram labeled by `generator_name`
- Add `roles/monitoring.metricWriter` IAM grant to `scripts/gcp_setup.sh`
- Add `GE_SERVICE_NAME` and `GE_METRICS_EXPORT_INTERVAL_SEC` to `.env.example`

# Testing

- `pipenv run pytest` — 352 tests passing (14 new tests for `MetricCollector` and `timed()` metric emission)
- Local: run with no `PROJECT_ID` set → metrics print to stdout on feed requests
- GCP integration: set `PROJECT_ID` and hit `/xrpc/app.bsky.feed.getFeedSkeleton`, verify `custom.googleapis.com/greenearth-api/feed.render.duration_ms` appears in Cloud Monitoring Metrics Explorer segmented by `feed_name`


src/app/lib/metrics.py
```diff
@@ -0,0 +1,138 @@
+"""OpenTelemetry-based metric collector for GCP Cloud Monitoring.
+
+Mirrors the OTelMetricCollector design from ingex (PR #231), ported to Python.
```
mjmor: We don't need to add this detail.

src/app/lib/metrics.py
```diff
@@ -0,0 +1,138 @@
...
+When GE_GCP_PROJECT_ID is set, metrics are exported to GCP Cloud Monitoring
+under the prefix ``custom.googleapis.com/greenearth-api/``.  When it is empty
+the stdout exporter is used, which is useful for local development.
+
+Instrument type is inferred from the metric name suffix, matching the ingex
+convention:
```
mjmor: We don't have to keep referencing the `ingex` convention–it's implicit. This applies to here and everywhere else in this PR.

src/app/lib/telemetry.py
```diff
@@ -9,27 +9,28 @@
...
-    stay quiet until someone is actively debugging performance. When a
-    request ID is set on the current ContextVar it is included as
-    ``rid=<id>`` so a single ``grep`` ties together all spans from one
-    request.
+    Timing logs are only emitted when ``GE_TIMING_LOGS=1`` is set.
+
+    When *metric_name* is provided and a MetricCollector is configured,
+    the elapsed milliseconds are recorded as a metric unconditionally —
+    independent of ``GE_TIMING_LOGS``.  Any keyword extras become metric
+    attributes (labels).
```
mjmor: I don't know that it makes sense to capture **all** keyword extras as metric attributes. Things like integer "count" (and other continuous variables) will have too many unique values to be useful as metric labels in the dashboard. What's a better way to handle this? I'm thinking: update `timed` function signature to have separate, named keyword extras; one used for metric labels and logging, and the other used just for logging.

src/app/lib/telemetry.py
```diff
@@ -9,27 +9,28 @@
 
 
 @asynccontextmanager
-async def timed(logger: logging.Logger, label: str, **extra: object):
-    """Async context manager that logs elapsed time for a labelled operation.
+async def timed(logger: logging.Logger, label: str, metric_name: str | None = None, **extra: object):
```
mjmor: Let's update the signature here: the `label` should be used as both the logging label and the metric_name. All places we currently log can be updated to provide a label that matches the metric format (e.g., `knn_search` become `es.knn.duration_ms`). Let's include a boolean field for `record_metric` which defaults to False to indicate whether to record the metric (instead of doing so implicitly on the presence of labels/names.

src/app/routers/xrpc.py
```diff
@@ -360,22 +295,16 @@ async def get_feed_skeleton(
             cached_uris = await feed_cache.retrieve(parsed.id)
         if cached_uris is not None:
             if parsed.offset < len(cached_uris):
-                # Serve from the existing cached batch.
```
mjmor, line 536: These changes are unrelated to the functional change in this PR. Please limit the scope of changes to what's relevant in this PR and don't include unnecessary stylistic changes that touch code not relevant to the targeted change.

This feedback applies to other places in the PR where unnecessary changes were made as well.

mjmor (commented):
Please address these comments.

src/app/lib/metrics.py
mjmor: Removed.

src/app/lib/metrics.py
mjmor: Removed throughout.

src/app/lib/telemetry.py
mjmor: Redesigned `timed()` to separate metric labels from log-only context: added `metric_attrs: dict[str, str] | None` for low-cardinality metric labels (e.g. `feed_name`, `generator_name`), while `**extra` remains for log-only values like counts and offsets. Neither bleeds into the other.

src/app/lib/telemetry.py
mjmor: `label` now doubles as the metric name. Added explicit `record_metric: bool = False` opt-in — metrics are only recorded when this is set. Instrumented call sites update their label to metric-format (e.g. `"feed.render.duration_ms"`); un-instrumented call sites (`feedcache_retrieve`, etc.) are untouched.

src/app/routers/xrpc.py
mjmor: Restored original `get_feed_skeleton` body intact (all original comments and logic preserved). The only change is an `async with timed(...)` wrapper around the serving block — necessary to capture all return paths — which requires one level of indentation on the existing code.

mjmor:
Expect @raindrift to have a relatively thorough review since you know how we did this in `ingex` repo (similar pattern here). Adding @davidfreifeld and @inthree3 for awareness (but welcome any feedback ofc).

PR #133, opened 2026-06-01
Don't use embeddings for posts with no usable content

### LINKED ISSUE #109: For posts with no usable content, do not trust their content embeddings
I think we get content embeddings, even for posts that have no truly usable content.

The embedding should be over the post content, alt text for media, and video transcript.

If none of those things are present on a post (like it's just an image), zero the embedding.

Also include a guard in the similarity search candidate generator that drops results if their embedding is empty. Empty embeddings are allowed for candidate generators that don't search on it, though.

davidfreifeld:
Closes #109 

We have some posts with embeddings but no usable content. We were told by Graze that embeddings could be based on `media.alt_text` or `video_transcript` - but in practice this does not seem to be true: posts without the `content` field but one of those other two (or `external_embed.description`) seem to have meaningless embeddings. So let's not use them as a result of post similarity candidate generation, and let's not pass those embeddings to the ranking stage. See below for the distribution of posts by the 5 fields. Thankfully, about 96% of posts have something in the `content` field and also have embeddings. 

Changes:
- The post similarity candidate generator now will no longer return posts without embeddings.
- For the other candidate generators (which use utils.py:candidate_posts_from_es_response() ), it will not attach an embedding if the `content` field is missing or empty.
- For fetch_post_embeddings() (used by post_similarity) and fetch_post_embeddings_and_authors() (used by two_tower), posts with embeddings but no `content` field will not be returned (those functions already do not return posts with missing embeddings). 

Screenshot 2026-06-02 at 2 22 14 PM

### DIFF
--- a/src/app/lib/candidates/post_similarity.py
+++ b/src/app/lib/candidates/post_similarity.py
@@ -111,8 +111,17 @@ async def knn_search_posts(
         candidates.append(candidate_post_from_hit(hit, generator_name=generator_name))
         if len(candidates) >= num_candidates:
             break
-
-    return candidates
+    
+    # drop candidates without embeddings
+    candidates_with_embeddings = [
+        candidate for candidate in candidates if candidate.minilm_l12_embedding
+    ]
+    if len(candidates_with_embeddings) < len(candidates):
+        logger.info(
+            "Dropped %d post-similarity candidates without embeddings",
+            len(candidates) - len(candidates_with_embeddings),
+        )
+    return candidates_with_embeddings
 
 
 class PostSimilarityCandidateGenerator(CandidateGenerator):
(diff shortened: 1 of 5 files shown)


src/app/lib/candidates/post_similarity.py
```diff
@@ -111,8 +111,17 @@ async def knn_search_posts(
         candidates.append(candidate_post_from_hit(hit, generator_name=generator_name))
         if len(candidates) >= num_candidates:
             break
-
-    return candidates
+    
+    # drop candidates without embeddings
```
raindrift, line 115: For now this is probably the way, though I think in the future we might want to add some kind of low-cardinality flag for whether the embedding is present, so that we can exclude it inside the es query instead.

raindrift (approved):

PR #137, opened 2026-06-03
Candidate generation only feeds for testing

### LINKED ISSUE #113: Make candidate generator-specific feeds
Create one hidden feed per candidate generator, with no infill, no ranker, and no diversification. So we can see the output of the generators on prod.

Depends on #99 which will enable us to hide these feeds from people searching in the feed store.

davidfreifeld:
Closes #113 

Create a new private feed for each of the candidate generators, basically for testing and debugging purposes. Each feed is just pure candidate generation with no ranking or diversification. The random candidate generator already has its own (public) feed, so I did not create a duplicate private one. 

Also renamed the "basic-similarity" to be "unranked-your-feed" because that is what it is - the "your-feed" feed, but without the ranker. 

Note that I did not change the rkey for the new "unranked-your-feed" feed, so that it will update in place when we deploy. I think the other option is to change it so it matches the name, and then delete the old feed when we deploy. Happy to do either one, so if you (kind reviewer) think the second option is better, let me know and I'll do that.

### DIFF
--- a/scripts/publish_feed_test.py
+++ b/scripts/publish_feed_test.py
@@ -214,7 +214,7 @@ class TestPublishFeed:
         record = put_call.kwargs["json"]["record"] if "json" in put_call.kwargs else put_call[1]["json"]["record"]
         assert record["$type"] == "app.bsky.feed.generator"
         assert record["did"] == GENERATOR_DID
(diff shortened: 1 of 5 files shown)


mjmor (approved):
LGTM. I think it's fine to leave the "unranked-your-feed" with the previous key since it has an accurate description on it in code.

PR #140, opened 2026-06-03
Suppress already-seen posts

### LINKED ISSUE #107: Omit posts the user has seen from ranking
Don't include seen posts in the ranking. It's probably ok to look only at posts seen in the last ~week, if we want to keep the size of that data small.

Depends on #106

raindrift:
We now ingest `sendInteractions` data, so we know which posts a user has actually viewed (`interactionSeen`). Previously a user could be re-served the same posts across feed loads — only within-batch repeats were avoided via the cursor. This excludes recently-seen posts from generation so the feed keeps moving forward.

### Approach

Reuses the existing `exclude_uris` mechanism that cursor pagination already relies on:

- **Persist** seen post URIs per user in Firestore, in daily buckets under `users/{did}/seen_posts/{YYYY-MM-DD}` (one array per UTC day). Appends use `ArrayUnion` (dedupes within a bucket); a native Firestore TTL on `expires_at` reaps buckets ~5 days out, mirroring the existing `feed_cache` TTL.
- **Write** happens in the existing background interactions task — `interactionSeen` items are collected per DID and recorded with no added request latency.
- **Read** happens on fresh feed generation (and regenerate-past-end-of-cache): the user's recent seen URIs are merged into `exclude_uris`. The read is fail-soft — a Firestore hiccup degrades the feature (possible repeats) rather than breaking feed serving.

### Tradeoffs (intentional)

- The exclude set is capped at the **1000 most-recent** seen URIs to bound the ES `must_not` payload; posts older than the cap may occasionally reappear. This is acceptable — accuracy is traded for performance here.
- The read filters on `expires_at > now`, which keeps reads bounded even before the TTL policy is enabled or while TTL deletion lags. (Discussed: the filter is essentially free at this scale and is the safety net that makes correctness independent of TTL being live.)

### Deploy note

Run `./scripts/gcp_setup.sh --environment stage` (and `prod`) once to enable the new `seen_posts` TTL policy. Until then buckets linger but are still filtered out at read time, so correctness does not depend on it.

### Incidental cleanup

Stopped the metrics collector from dumping a large `resource_metrics` JSON blob to stdout in local/dev. The console exporter is replaced with no exporter in non-deployed environments (metrics are still recorded; stage/prod export to Cloud Monitoring unchanged). Added a regression test asserting dev produces no stdout output.

Closes #107

### DIFF
--- a/scripts/gcp_setup.sh
+++ b/scripts/gcp_setup.sh
@@ -184,6 +184,21 @@ ensure_feed_cache_ttl_policy() {
         || log_warn "TTL policy may already exist or could not be updated (non-fatal)"
 }
 
+ensure_seen_posts_ttl_policy() {
+    local firestore_db
+    firestore_db="$(get_firestore_database)"
+
+    log_info "Ensuring TTL policy on seen_posts.expires_at..."
+    gcloud firestore fields ttls update expires_at \
+        --collection-group=seen_posts \
+        --database="$firestore_db" \
+        --project="$PROJECT_ID" \
+        --enable-ttl \
+        --quiet 2>/dev/null \
+        && log_info "TTL policy enabled on seen_posts.expires_at" \
+        || log_warn "TTL policy may already exist or could not be updated (non-fatal)"
+}
+
 ensure_firestore_api_key_secret() {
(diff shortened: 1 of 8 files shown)


mjmor (approved):
LGTM! Thanks for cleaning up the local metric logging. I was using that to test and forgot to drop the code.

PR #132, opened 2026-05-31
Post-ranking with the Perspective API

max-ai-ast:
Closes #111

This PR adds Perspective API scoring to the feed ranking pipeline. Perspective runs as one of an arbitrary number of weighted **rank models** alongside the two-tower model. Each model's raw scores are normalized into a common `[-1, 1]` range using its theoretical bounds, then combined into a single ranking via a weighted average using per-model relative weights configured per feed. This generalizes the existing weighted-generator pattern (`GeneratorSpec`) to ranking.

## Formula

The PRC score is now computed as the exact weighted sum from `perspective_baseline_minus_outrage_toxic` in the [PRC reference implementation](https://github.com/HumanCompatibleAI/ranking-challenge-perspective/blob/main/perspective_ranker.py#L163-L179) — the "Uprank Bridging, Downrank Toxic" condition:

```
+ 1/6  each: REASONING, PERSONAL_STORY, AFFINITY, COMPASSION, RESPECT, CURIOSITY (_EXPERIMENTAL)
- 1/6  each: FEARMONGERING, GENERALIZATION (_EXPERIMENTAL)
- 1/18 each: SCAPEGOATING, MORAL_OUTRAGE, ALIENATION (_EXPERIMENTAL)
- 1/8  each: TOXICITY, IDENTITY_ATTACK, INSULT, THREAT
```

These weights conveniently sum to exactly `(-1.0, +1.0)`, matching the `[-1, 1]` normalization range the new rank-model framework expects — so Perspective's theoretical `score_bounds` require no rescaling beyond a float-precision clamp.

# This PR

- **`src/app/models.py`**: Added `RankModelSpec` (mirrors `GeneratorSpec`'s name+weight pattern) and replaced `RankPredictRequest.model: str | None` with `models: list[RankModelSpec]` (required, extensible to any number of weighted models). Removed `FeedConfig.use_perspective` — Perspective participation is now controlled solely by whether `models` includes a `perspective` entry.
- **`src/app/lib/rankers/base.py`**: Added an abstract `score_bounds: tuple[float, float]` property to the `Ranker` interface — the theoretical (min, max) range of a ranker's raw scores, used to linearly normalize each model's output into `[-1, 1]` before combination.
[...]

PR #145, opened 2026-06-05
Collect and show feed debug data

### LINKED ISSUE #114: Capture feed debugging information
We want a way to see the details of how a feed item was chosen, after seeing it in a feed.

We should store this data in firebase. For each item we want to know the ranking scores, which candidate generator(s) returned it, where it ended up in the final order after diversification, etc. Basically, we want to be able to answer the question "why this item?" so we want to know how its ranking changed as it moved through the pipeline. 

We probably also want to know what the input parameters were for that feed load. Like, which liked posts were used to compute search vectors, what data was used to assemble the user embedding, etc. If items were discarded, we should know about those too.

We also need a way to find this data for a given feed load once it has run. This can be a simple command-line tool to start. For the tool, I think it would be nice to supply a username and be shown a list of recent feed loads. Then I could pick a feed load, and see the output for it.

It probably makes sense to put this behind a feature flag or something, so that it is only enabled for specific users or groups of users, as I expect it to have a performance impact. I don't think it's time for a feature flag library yet, so let's just add a flag to the user record for now.

raindrift:
Closes #114.

When a post shows up in a feed, we currently have no way to answer "why this item?" after the fact. The retrieve → rank → diversify pipeline discards all intermediate signal: which generator produced a candidate, its scores at each stage, how the ranker reordered it, and where diversification finally placed it. This PR captures that information per feed load so we can inspect it later, and adds a CLI to do so.

### Approach

Capture is driven by a `ContextVar` recorder (`lib/feed_debug.py`), mirroring the existing `request_cache_scope` pattern. Pipeline stages call record helpers that are no-ops unless a recorder is installed, so there is **zero cost unless a debug-enabled user triggers it**. The recorder holds the real pipeline objects (`CandidateGenerateRequest`, `CandidateResult`, `RankPredictResult`, `CandidatePost`) rather than parallel debug types, so it keeps capturing new fields as the ranking pipeline evolves; the per-item "why this item?" view is assembled at display time by the CLI.

Capture is gated behind a per-user `debug_feeds` flag on the user record (no feature-flag library yet, per the issue). The flag is read once on the feed hot path (fail-soft); when enabled, the pipeline runs under a recorder scope and the assembled document is written in a background task — handle resolution and the Firestore write stay off the serving path.

### What's included

- **Per-user opt-in flag** `debug_feeds` on `UserDocument`.
- **Debug records** stored at `users/{did}/feed_debug/{request_id}`, one per feed load, with a 7-day native TTL (`expires_at`). TTL policy registered in `scripts/gcp_setup.sh` alongside the existing `feed_cache`/`seen_posts` policies.
- **`CandidatePost` enrichment**: added `author_username` plus media metadata (`contains_images`, `contains_video`, `image_count`, `video_count`, `external_uri`). Media comes from Elasticsearch via the shared `candidate_post_from_hit`; author handles are resolved DID→handle via the existing id resolver in the background writer. These are canonical post attributes, not debug-only fields.
- **Instrumentation** of candidate generation, the `post_similarity`/`two_tower` user-feature steps, and the rank/diversify stages — no signature changes.
- **CLI** `scripts/feed_debug.py`: `feed_debug <user> --enable|--disable|--list|--show <request_id>`, with `rich`-rendered output that shows each item's journey (`retrieved by <generator> (gen score) → ranked #N (model score) → final #pos`), media, snippet, and a discarded-candidates table. `--environment dev|stage|prod` (default `dev`/emulator; `--env` alias) selects the Firestore target.

### Notes

- `rich` added to `[dev-packages]` (the CLI is a local debug tool, never imported by the server).
- Embeddings are stripped and post content is truncated before storage to keep documents well under Firestore's 1 MB limit.
- The `feed_debug` TTL policy must be applied to stage/prod by running `scripts/gcp_setup.sh` against each (the local emulator ignores TTL policies).
- Tested: recorder/build-document, document + Firestore-helper, media-field, and end-to-end xrpc capture tests; full suite green.

### DIFF
--- a/src/app/routers/xrpc.py
+++ b/src/app/routers/xrpc.py
@@ -58,6 +58,7 @@ router = APIRouter(tags=["xrpc"])
 # Configuration
 # ---------------------------------------------------------------------------
 
+
 def _get_service_did() -> str:
     """Return the DID of this feed generator service.
 
@@ -73,7 +74,7 @@ def _get_hostname() -> str:
     did = _get_service_did()
     # did:web:<hostname> → hostname
     if did.startswith("did:web:"):
-        return did[len("did:web:"):]
+        return did[len("did:web:") :]
     return "localhost"
 
 
@@ -110,6 +111,7 @@ async def _resolve_username(request: Request, user_did: str) -> str:
 # Response models
 # ---------------------------------------------------------------------------
 
+
 class FeedLink(BaseModel):
     uri: str = Field(..., description="AT URI of the feed")
 
@@ -134,6 +136,7 @@ class FeedSkeletonResponse(BaseModel):
     When ``cursor`` is ``None`` it is omitted from the JSON output — the
     AT Protocol spec requires the field to be absent rather than ``null``.
     """
+
     model_config = {"populate_by_name": True}
 
     feed: list[SkeletonItem] = Field(default_factory=list)
@@ -177,7 +180,9 @@ class Interaction(BaseModel):
     model_config = {"populate_by_name": True}
 
     item: str | None = Field(default=None, description="AT URI of the post interacted with")
-    event: str | None = Field(default=None, description="Interaction event type (app.bsky.feed.defs#...)")
+    event: str | None = Field(
+        default=None, description="Interaction event type (app.bsky.feed.defs#...)"
+    )
     feed_context: str | None = Field(
         default=None,
         validation_alias="feedContext",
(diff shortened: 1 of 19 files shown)


src/app/routers/xrpc.py
```diff
@@ -399,6 +450,58 @@ async def _record_session(request: Request, user_did: str, feed_name: str, db) -
...
+    request: Request,
+    db,
+    recorder: FeedDebugRecorder,
+    request_id: str,
+    user_did: str,
+    generated_at: datetime,
+) -> None:
+    """Resolve author handles, assemble the debug document, and persist it.
+
+    Runs as a background task so handle resolution and the Firestore write stay
```
mjmor: Looks like the caller spawns the background task, yet the docstring makes it seem like this function guarantees a background task is spawned. Let's update to clarify so future LLMs are not confused.

mjmor (approved):
Nice! 🥇 

We might also want to capture MMR scores to understand why the final rank was different (what portion was due to author diversity vs. content diversity), but we can do that in a follow up.

I'll also need to write in perspective API impacts on final rank order with #132 .

PR #146, opened 2026-06-06
Use colon-free user document IDs

raindrift:
> **Base branch:** `raindrift/feed-debug-data` (stacked — review/merge that first). Diff is scoped to the document-ID change.

User documents are keyed by the full DID (`did:plc:<id>`). Colons in a Firestore document ID break subcollection navigation in the **emulator UI** — there's no clickable link from a user doc to its subcollections (`feed_debug`, `feed_activity`, `seen_posts`), so they appear missing even though the data is there and reachable by direct URL. This makes the new feed-debug data effectively un-browsable in local dev.

### Change

User documents are now keyed by the DID with the `did:plc:` prefix stripped (e.g. `did:plc:abc123` → `abc123`), removing colons from the key. The full DID is still stored in the document's `user_did` field — only the document *key* changes. Other DID methods pass through unchanged (all current users are `did:plc`).

This is centralized in `lib/firestore.py`: a single `user_doc_id()` helper, applied at every user-document key site. Callers continue to pass full DIDs; the mapping happens at the Firestore boundary, so no other code or the CLI changes.

### Impact — clean reset, no migration

Existing user docs keyed by the full DID are abandoned; new requests create fresh docs under the stripped key. We're treating this as an intentional reset rather than migrating:

- Local dev: just regenerate (re-enable the flag, load feeds).
- Stage/prod: resets the `users` collection — `created_at` history is lost and seen-post exclusions reset (the latter self-heals; users may briefly re-see recent posts). No data of lasting value is dropped.

### DIFF
--- a/src/app/lib/firestore.py
+++ b/src/app/lib/firestore.py
@@ -34,6 +34,18 @@ SEEN_POSTS_RETENTION_DAYS = 5
 # How long a feed-debug record lives before native Firestore TTL deletes it.
 FEED_DEBUG_RETENTION_DAYS = 7
 
+# Prefix stripped from a DID to form the user document ID. The full DID is
+# still stored in the document's ``user_did`` field; only the document *key* is
+# shortened. This keeps colons out of the key — colons in a document ID break
+# subcollection navigation in the Firestore emulator UI. All users are
+# currently did:plc; other DID methods are passed through unchanged.
+_USER_DID_PREFIX = "did:plc:"
+
+
+def user_doc_id(user_did: str) -> str:
+    """Map a DID to its Firestore user-document ID (colon-free for did:plc)."""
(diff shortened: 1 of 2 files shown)


raindrift:
After we deploy this to prod, it'll be good to remove the stale user records.

mjmor (approved):
Lgtm!

Just one point to make sure I'm understanding correctly: I don't see any read-path transformations back to at_uri format, so I guess we don't programmatically read any of this data today where we would need it in at_uri format, correct?

P.S. I changed the based branch to the debug feed branch because it made the PR additive and easier to review.

raindrift:
Yes, AFAIK we don't read the user_did from firebase at the moment. When accessing it we always have a request, so we use the DID from the request. We mostly only read subcollections from the user record, rather than using the user document itself for anything.

PR #152, opened 2026-06-10
Better Handling of Missing Rank Candidate Scores

### LINKED ISSUE #151: Decide on no-embedding rank score fallback
In #132 we introduced logic to combine multiple ranker scores. So we normalize them to between -1 and 1. We also replace rank_score=None with 0.0. The two tower model appends candidates without embeddings at the end and gives them rank_score=None. Those posts will now be treated as "neutral" by the rank score combiner. This may or may not be what we want. We might want to investigate what this looks like for real posts. We might decide to give those posts a score of -1, or exclude them from the result, for example.

Decision:
Let's set it to the request-level median, per ranker, based on the results that did have scores.

If no rankers have a usable value, remove it from the result, but also log a metric saying we did that so we know how much stuff we are dropping

davidfreifeld:
Closes #151 

Some changes to how we handle missing scores in the ranking stage (which now does a weighted average of one or more rankers, like two_tower and perspective): 
- If a candidate has no valid score from any of the rank models, it is excluded from the results.
  - Logs the number of candidates dropped because of this. Uses the metric collector. Not sure if I did this in the correct way - I didn't see a good example to go off of so I did my best.
- If a candidate has a valid score from at least one rank model, then we fill in any missing scores with the (normalized) median score for that model's results.
- If a given rank model has no valid scores at all, for any of the results, ignore that model and just use the other ones.

### DIFF
--- a/src/app/lib/rankers/predict.py
+++ b/src/app/lib/rankers/predict.py
@@ -151,21 +151,21 @@ async def run_predict(
                 normalized_by_model[model_name][uri] = normalized_score
 
     metric_collector = get_metric_collector()
-    if metric_collector:
+    if metric_collector and dropped_candidate_count > 0:
         metric_collector.record(
             "rank.predict.dropped_candidates_count",
             dropped_candidate_count,
         )
 
-    weights_by_model: dict[str, float] = {
-        name: weight
-        for name, weight, _ in models_with_valid_results
-    }
     if rec is not None:
-        for model_name, scores_dict in normalized_by_model.items():
-            rec.record_model_scores(model_name, weights_by_model[model_name], scores_dict)
+        for model_name, weight, _ in resolved:
+            rec.record_model_scores(
+                model_name,
+                weight,
+                normalized_by_model.get(model_name, {}),
+            )
 
-    total_weight = sum(weights_by_model.values())
+    total_weight = sum(weight for _name, weight, _ranker in models_with_valid_results)
 
     def _combined(at_uri: str) -> float:
         return sum(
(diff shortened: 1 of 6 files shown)


src/app/lib/rankers/predict.py
```diff
@@ -101,40 +109,89 @@ async def run_predict(
...
     for (name, weight, ranker), result in zip(resolved, results):
-        bounds = ranker.score_bounds
+        # all the valid uris with their scores from this ranker model
         raw_by_uri = {
-            ranking.at_uri: ranking.rank_score
-            for ranking in result.result.rankings
-            if ranking.at_uri
+            r.at_uri: r.rank_score
+            for r in result.result.rankings
+            if r.at_uri is not None and r.rank_score is not None
```
mjmor, line 121: Am I understanding correctly that only candidates with `score=None` will be considered as "invalid" scores that are later set to the request-level median? 

If so, we need to update the perspective client to properly use `None` as the sentinel value. Currently it uses `0.0` in a few places on failure / lack of signal instead:
- https://github.com/greenearth-social/api/blob/5c757cb89913c1302b573905f5de7b06a17035a9/src/app/lib/rankers/perspective.py#L43
- Multiple places in this method: https://github.com/greenearth-social/api/blob/5c757cb89913c1302b573905f5de7b06a17035a9/src/app/lib/perspective.py#L182

mjmor (approved):
LGTM (including the metric recording usage). Just one comment re: perspective API possibly needing to be updated to conform to the behavior expected here.

src/app/lib/rankers/predict.py
davidfreifeld: Great catch, thank you. I just updated the perspective ranker to use `None` as the sentinel value, which I think makes sense.

PR #153, opened 2026-06-10
Two Tower Candidate Generator

### LINKED ISSUE #108: Two-tower candidate generator
Make a candidate generator that will use the user tower to compute a search vector, then perform an ANN search in ES for that search vector.

The work to store the post embeddings is in [issue 372](https://github.com/greenearth-social/ingex/issues/372). The work for this story should be done after that one.

The current user tower of the two tower model accepts a user's recently liked posts. For each post, the features are the post's content embeddings and its author DID. There is already code to run inference via the inference-service on the user tower with these inputs in the two_tower ranker in the API repo. So this code should be mimicked/copied into the candidate generator.

For now, leave the two_tower ranker in place. Using the same model for candidate generation and ranking may seem silly, but in fact the ranker will be able to correct for the approximate-ness of the ANN search in the candidate generation phase. We will improve the ranking model in future work.

Note that there is a maximum batch size parameter in inference-service, that was set somewhat arbitrarily. We might want to increase this and experiment with the batch sizes we send to inference-service for performance tuning.

davidfreifeld:
Closes #108 

## This PR
Implements the two tower model as a candidate generator.

The post model UUID is retrieved from the inference service's /ready endpoint. Then, on a two tower candidate generation request, we run the user tower (just like in the two tower ranker). Then we query ES for the most similar posts based on the new generated post embeddings present on the posts. We filter to only posts that have the matching post model UUID. 

I refactored the code a bit because there was a lot of code we could reuse from post_similarity and the ranker version of two_tower. So I moved functions into shared files to be used between the modules. 

### Post-tower UUID lookup, caching, and error handling
The two-tower candidate generator now asks the inference-service `/ready` endpoint for the currently configured post-tower model UUID before running the ES kNN search. That UUID is passed into the ES query so we only retrieve posts whose `ge_post_embedding_model_uuid` matches the model that produced the user embedding.

The `/ready` response handling distinguishes between two cases:

- If `/ready` returns a valid response with no `post-tower` model entry, that means post-tower is not configured. The two-tower candidate generator logs a warning and returns no candidates.
- If `/ready` returns a malformed response, for example missing `models`, malformed model entries, or a `post-tower` entry without a non-empty `model_uuid`, that is treated as an inference-service contract error and raises `InferenceResponseFormatError`.

The post-tower UUID is cached process-wide by `(base_url, api_key)` with a 5-minute TTL. Refreshes use a per-key `asyncio.Lock`, so concurrent requests collapse into a single `/ready` call instead of stampeding the inference-service.

There is also a stale-if-error grace window. If the cached UUID has expired but is less than 1 hour past expiry, and the refresh request fails with an exception or HTTP/format error, we log a warning and keep using the last known UUID for that request. This handles transient inference-service `/ready` failures without immediately disabling two-tower candidates. A successful `/ready` response with “post tower not configured” returns `None` for the current request instead of falling back to stale, so intentional unconfiguration is respected.

## Testing
- Tested candidate generation endpoint with my DID - got relevant posts!
- Deployed to stage, successfully loaded Your Feed
- Test coverage for all changes; pytest runs successfully


src/app/lib/candidates/es_candidates.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+# How many extra hits to fetch from ES per requested candidate. The kNN
+# neighborhood of an averaged-likes vector is empirically ~75% replies and
+# ~5-10% videos, so we need significant margin to still hit num_candidates
+# after Python-side filtering. Capped so we don't blow out latency on large
+# requests.
+OVERFETCH_MULTIPLIER = 5
+MIN_OVERFETCH = 60
+MAX_OVERFETCH = 500
```
mjmor: Sorry, you're going to have merge conflicts with : https://github.com/greenearth-social/api/pull/168

src/app/lib/candidates/two_tower.py
```diff
@@ -0,0 +1,74 @@
...
+            # run the user tower to get the user embedding
+            user_embedding = await compute_user_embedding(
+                user_did,
+                es,
+                inference_base_url,
+                inference_api_key,
+                TWO_TOWER_GENERATOR_NAME,
+            )
+
+        async with timed(logger, "two_tower_candidate_posts_search", n_candidates=num_candidates):
```
mjmor: Do we need this and the previous `timed` calls if the `knn_search` method already has a `timed` call internally? Frankly, not too picky either way for me, just making sure you're aware. Let's keep moving forward if that was an intentional decision.

src/app/feeds.py
```diff
@@ -62,7 +62,7 @@
         internal_display_name="a0 YF",
         gen_request_template=CandidateGenerateRequest.model_construct(
             generators=[
-                GeneratorSpec(name="post_similarity", weight=0.5),
+                GeneratorSpec(name="two_tower", weight=0.5),
```
mjmor, line 65: yesss!!!

mjmor (approved):
Looks good - let's do it!

src/app/lib/candidates/two_tower.py
davidfreifeld: Good point - removed both of them!

PR #166, opened 2026-06-19
Add scheduled feed probes for continuous latency monitoring

### LINKED ISSUE #165: Add scheduled feed probe for consistent latency data
## Background

`feed.render.duration_ms` is already exported to Cloud Monitoring, but data points are sporadic since we are still in pilot and don't have consistent traffic. This makes it hard to track service performance and do load testing.

## What to do

Set up a Cloud Scheduler job that hits the feed endpoint once per minute.
Cloud Scheduler can call the URL directly, so no Cloud Run Job needed.

Endpoint to probe (your-feed, small limit to keep it cheap):
```
GET https://greeneartprod-21637448064.us-east1.run.app/xrpc/app.bsky.feed.getFeedSkeleton?feed=at://did:plc:wrmpulygwvuhjn2c3jbalgqj/app.bsky.feed.generator/your-feed&limit=5
```

  Schedule: `* * * * *` (every 1 minute), both stage and prod.

  ## Why this matters

  With consistent traffic we can:
  - See a continuous latency trend in Cloud Monitoring instead of sparse dots
  - Confirm that kNN / index changes from #148 have actually improved things

  ## Notes

  - Document the scheduler setup in `api/scripts/gcp_setup.sh` alongside the existing
    `setup_expiry_cloud_scheduler` pattern in ingex for consistency
  - No auth needed since the feed endpoint is public

inthree3:
## Summary

- Adds Cloud Scheduler jobs that probe all three public feeds (`random`, `your-feed`, `best-of-friends`) every minute, so `feed.render.duration_ms` gets a continuous signal in Cloud Monitoring instead of sparse dots.
- AT Protocol JWTs expire in ~20 minutes and cannot be pre-issued for a scheduler, so a `probe-secret-{stage,prod}` is generated in Secret Manager and validated via `X-Probe-Secret` header in `getFeedSkeleton` to bypass AT Protocol auth for probe requests.
- Probe DID is `greenearth-social.bsky.social` (`did:plc:wrmpulygwvuhjn2c3jbalgqj`), the project's operational account. This allows user-specific generators (`followed_users`, `post_similarity`) to run without errors. `seen_posts` accumulation is intentionally not bypassed, because the account is operational so there is no UX impact, and adding a bypass would introduce unnecessary complexity.

## Results (stage)
Now the rendering duration is constantly monitored and can be seen in the dashboard :)
image

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -208,6 +208,7 @@ deploy_api_service() {
     local inference_api_key_secret="inference-api-key-stage"
     local firestore_api_key_secret="firestore-api-key-stage"
     local feed_context_secret="feed-context-secret-stage"
+    local probe_secret="probe-secret-stage"
     local perspective_api_key_secret="perspective-api-key-stage"
     local firestore_database="greenearth-stage"
     if [ "$ENVIRONMENT" = "prod" ]; then
@@ -215,6 +216,7 @@ deploy_api_service() {
         inference_api_key_secret="inference-api-key-prod"
         firestore_api_key_secret="firestore-api-key-prod"
         feed_context_secret="feed-context-secret-prod"
+        probe_secret="probe-secret-prod"
         perspective_api_key_secret="perspective-api-key-prod"
         firestore_database="greenearth-prod"
     fi
@@ -248,6 +250,7 @@ deploy_api_service() {
     deploy_cmd="$deploy_cmd --set-secrets=GE_INFERENCE_API_KEY=$inference_api_key_secret:latest"
     deploy_cmd="$deploy_cmd --set-secrets=GE_FIRESTORE_API_KEY=$firestore_api_key_secret:latest"
     deploy_cmd="$deploy_cmd --set-secrets=GE_FEED_CONTEXT_SECRET=$feed_context_secret:latest"
+    deploy_cmd="$deploy_cmd --set-secrets=GE_PROBE_SECRET=$probe_secret:latest"
     deploy_cmd="$deploy_cmd --set-secrets=GE_PERSPECTIVE_API_KEY=$perspective_api_key_secret:latest"
 
     # Resource and scaling configuration
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -240,6 +240,7 @@ deploy_api_service() {
     deploy_cmd="$deploy_cmd --set-env-vars=GE_ELASTICSEARCH_VERIFY_SSL=false"
     deploy_cmd="$deploy_cmd --set-env-vars=GE_FIRESTORE_PROJECT=$PROJECT_ID"
     deploy_cmd="$deploy_cmd --set-env-vars=GE_FIRESTORE_DATABASE=$firestore_database"
+    deploy_cmd="$deploy_cmd --set-env-vars=GE_PROBE_USER_DID=did:plc:wrmpulygwvuhjn2c3jbalgqj"
 
     if [ -n "$GE_INFERENCE_BASE_URL" ]; then
         deploy_cmd="$deploy_cmd --set-env-vars=GE_INFERENCE_BASE_URL=$GE_INFERENCE_BASE_URL"
--- a/src/app/routers/xrpc.py
+++ b/src/app/routers/xrpc.py
@@ -679,7 +679,8 @@ async def get_feed_skeleton(
         logger.error("Firestore client not initialized")
         raise HTTPException(status_code=500, detail="Firestore unavailable")
 
-    _spawn_background(_record_session(request, user_did, feed_name, db))
+    if user_did != "did:plc:probe":
+        _spawn_background(_record_session(request, user_did, feed_name, db))
 
     # Per-user opt-in: capture pipeline debugging info for this feed load. This
     # costs one extra Firestore read per request; fail-soft so a hiccup degrades
--- a/src/app/routers/xrpc.py
+++ b/src/app/routers/xrpc.py
@@ -652,20 +652,25 @@ async def get_feed_skeleton(
 
     feed_cfg = FEEDS[feed_name]
 
-    # Authenticate the requesting user via the AT Protocol inter-service JWT.
-    # A valid DID is required for this feed endpoint.
-    user_did = await verify_auth_header(request, service_did=_get_service_did())
+    # Cloud Scheduler probe bypass: if GE_PROBE_SECRET is set and the request
+    # carries the matching X-Probe-Secret header, skip AT Protocol auth and
+    # use a synthetic DID so the full pipeline runs and emits latency metrics.
+    _probe_secret = os.environ.get("GE_PROBE_SECRET")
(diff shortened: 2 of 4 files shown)


src/app/routers/xrpc.py
```diff
@@ -652,20 +653,27 @@ async def get_feed_skeleton(
...
-            headers={"WWW-Authenticate": "Bearer"},
-        )
+    # Cloud Scheduler probe bypass: if GE_PROBE_SECRET is set and the request
+    # carries the matching X-Probe-Secret header, skip AT Protocol auth and
+    # use the configured probe DID so the full pipeline runs and emits latency metrics.
+    _probe_secret = os.environ.get("GE_PROBE_SECRET")
+    if _probe_secret and hmac.compare_digest(
+        request.headers.get("X-Probe-Secret", ""), _probe_secret
+    ):
+        user_did = os.environ.get("GE_PROBE_USER_DID", "did:plc:wrmpulygwvuhjn2c3jbalgqj")
```
mjmor: This is probably fine for now, but this means the probes are going to be generating data in our production databases, like feed view timestamps and the last time we saw the user, etc. We don't need to address it in this PR. We can always filter out our account when analyzing that data, but would you file a follow up issue (probably post MVP priority) to implement a bypass that skips writing to production databases for the probe requests and also delete all probe data from the production databases after that issue is complete?

mjmor (approved):
yes!!!! :)

scripts/deploy.sh
```diff
@@ -238,6 +240,7 @@ deploy_api_service() {
     deploy_cmd="$deploy_cmd --set-env-vars=GE_ELASTICSEARCH_VERIFY_SSL=false"
     deploy_cmd="$deploy_cmd --set-env-vars=GE_FIRESTORE_PROJECT=$PROJECT_ID"
     deploy_cmd="$deploy_cmd --set-env-vars=GE_FIRESTORE_DATABASE=$firestore_database"
+    deploy_cmd="$deploy_cmd --set-env-vars=GE_PROBE_USER_DID=did:plc:wrmpulygwvuhjn2c3jbalgqj"
```
mjmor, line 243: Oh!! Also, would you please use our testing account for probes before deploy/merge? https://bsky.app/profile/caterpie-internal.bsky.social

You'll have to lookup the DID for the account. I don't have it off hand.

mjmor:
@inthree3 , I'm also seeing some authentication error logs: 

```
{
  "textPayload": "INFO:     169.254.169.126:39942 - \"GET /xrpc/app.bsky.feed.getFeedSkeleton?feed=at://did:plc:wrmpulygwvuhjn2c3jbalgqj/app.bsky.feed.generator/best-of-friends&limit=30 HTTP/1.1\" 401 Unauthorized",
  "insertId": "6a349b9a000ca2a1aed37630",
  "resource": {
    "type": "cloud_run_revision",
    "labels": {
      "project_id": "greenearth-471522",
      "configuration_name": "greenearth-api-stage",
      "service_name": "greenearth-api-stage",
      "location": "us-east1",
      "revision_name": "greenearth-api-stage-00121-wgt"
    }
  },
  "timestamp": "2026-06-19T01:30:02.828065Z",
  "labels": {
    "instanceId": "002f8ffd47611731ee955352a90e028e60340644b8a37a8c36b2e380bfaca66608016fa1aee5c3fb83525942b791ba8492c108b8352e93aaa03592cea52df11bf6e427edbd2445cee9fa3487d3"
  },
  "logName": "projects/greenearth-471522/logs/run.googleapis.com%2Fstdout",
  "receiveTimestamp": "2026-06-19T01:30:03.139143422Z"
}
```

Could you confirm these are not caused by the probe before merge / deploy? They may be false noise (since we're bypassing auth) but if so, we should make sure the probe bypasses the 401 error logging too if the probe is the requestor to keep the logs clean for future debugging.

inthree3:
> @inthree3 , I'm also seeing some authentication error logs:
> 
> ```
> {
>   "textPayload": "INFO:     169.254.169.126:39942 - \"GET /xrpc/app.bsky.feed.getFeedSkeleton?feed=at://did:plc:wrmpulygwvuhjn2c3jbalgqj/app.bsky.feed.generator/best-of-friends&limit=30 HTTP/1.1\" 401 Unauthorized",
>   "insertId": "6a349b9a000ca2a1aed37630",
>   "resource": {
>     "type": "cloud_run_revision",
>     "labels": {
>       "project_id": "greenearth-471522",
>       "configuration_name": "greenearth-api-stage",
>       "service_name": "greenearth-api-stage",
>       "location": "us-east1",
>       "revision_name": "greenearth-api-stage-00121-wgt"
>     }
>   },
>   "timestamp": "2026-06-19T01:30:02.828065Z",
>   "labels": {
>     "instanceId": "002f8ffd47611731ee955352a90e028e60340644b8a37a8c36b2e380bfaca66608016fa1aee5c3fb83525942b791ba8492c108b8352e93aaa03592cea52df11bf6e427edbd2445cee9fa3487d3"
>   },
>   "logName": "projects/greenearth-471522/logs/run.googleapis.com%2Fstdout",
>   "receiveTimestamp": "2026-06-19T01:30:03.139143422Z"
> }
> ```
> 
> Could you confirm these are not caused by the probe before merge / deploy? They may be false noise (since we're bypassing auth) but if so, we should make sure the probe bypasses the 401 error logging too if the probe is the requestor to keep the logs clean for future debugging.

I found that the error log was caused by a deployment made before merging this branch, which removed `GE_PROBE_SECRET` and caused a 401 error.

If we deploy without the line below, an auth error occurs:
`--set-secrets=GE_PROBE_SECRET=probe-secret-stage:latest`

After merging this branch and redeploying, the issue should be fixed.

scripts/deploy.sh
inthree3: Fixed it!

src/app/routers/xrpc.py
inthree3: Sure, I'll write the issue! By the way, I think bypassing the read/write for the probe user is not urgent, since we don't use other users' history for feed generation, and the probe user is a test account.

My bigger concern at the moment is that the probe duration doesn't seem to be similar to the real feed generation latency. I think it's generally faster than the real user's. I need to examine more to confirm this problem exists, but if it's true, I guess the steady user engagement history and its following network is the cause.

I asked Claude and it let me know what user history is needed for each feed generation, and that inference calculation doesn't actually operate if the user's like history doesn't exist. It also suggests that the history doesn't have to be dynamically changed, since whether the inference calculation is operating is more important.

I'd like to add this point to the probe improvement issue!
@mjmor Do you have thoughts on this?

mjmor: I agree bypassing writes for the probe user is not urgent. 

I also agree about the probe account not accurately capturing the latency of real users. I started using the Caterpie test account (following accounts, liking posts, etc.) earlier today to make it more realistic. Feel free to do the same! 

What is the alternative solution / follow up issue you're imagining though?

PR #194, opened 2026-06-25
Tag all request-path metrics with their originating endpoint

### LINKED ISSUE #192: Tag logged metrics with the api endpoint that generated them
We have some people who might use our candidate generator API for a project they're working on, which is great.

However, we need a way to differentiate the metrics we log for those requests from the ones that are for specific feeds we are serving.

I think the best way to do this is to tag all logged metrics with the endpoint that they were generated from. That way we can differentiate which requests/queries/etc came from `getCandidates` vs `getFeedSkeleton`, and we can know if the load from our api users is having an impact on overall performance.

I think it should be possible to include this in the metrics logging library, such that we get it on every logged metric without having to update every callsite.

For now let's do the easy thing and constrain this to the api service, where the endpoint is known. If we end up needing it later we can pass it through to requests from the inference service, etc, but for now let's forego the complexity.

raindrift:
Closes #192 

Metrics recorded during request handling had no way to distinguish which API endpoint produced them — a `feed.render.duration_ms` from `getFeedSkeleton` looked identical to one from `candidates/generate`, making per-endpoint breakdowns in GCP impossible without threading an endpoint name through every `record()` callsite.

This adds an `endpoint` label automatically. The HTTP middleware resolves the matched route name up front and stores it in an `_endpoint` ContextVar (alongside the existing request-id ContextVar), and `MetricCollector.record` reads it so every metric on the request path is tagged for free, including metrics from background tasks spawned by the handler.

### Notes
- Route matching happens before `call_next` so the value is available throughout request handling. Requests that don't fully match a named route (e.g. 404s) get no label, avoiding a high-cardinality `endpoint` value.
- An explicitly passed `endpoint` attribute always wins over the ContextVar.

### DIFF
--- a/src/app/lib/metrics.py
+++ b/src/app/lib/metrics.py
@@ -97,8 +97,20 @@ class MetricCollector:
 
         Instruments are lazily created based on the name suffix (see module
         docstring).  Attributes become GCP metric labels.
+
+        When recorded on a request path, the metric is automatically tagged
+        with an ``endpoint`` label naming the originating API endpoint (e.g.
+        ``get_feed_skeleton`` vs ``candidates_generate``). This is read from a
+        ContextVar set by the endpoint middleware, so every callsite gets it
+        for free. An explicitly passed ``endpoint`` attribute wins.
         """
-        attrs = dict(attributes) if attributes else None
+        from .request_context import get_endpoint
+
+        labels: dict[str, str] = dict(attributes)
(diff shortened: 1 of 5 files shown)


mjmor (approved):
Lgtm thanks for getting ahead of this

inthree3 (approved):
LGTM! Happy to approve as-is.

Issue #196, opened 2026-06-30
Grant Cloud Monitoring metricDescriptors.create permission to API service account


mjmor:
## Problem

The `greenearth-api-prod` Cloud Run service's runtime service account lacks the `monitoring.metricDescriptors.create` IAM permission. The OpenTelemetry Cloud Monitoring exporter (`src/app/lib/metrics.py`, `MetricCollector`, used when `ENVIRONMENT` is `stage`/`prod`) fails on every periodic export interval (`GE_METRICS_EXPORT_INTERVAL_SEC`, default 60s) with:

```
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.PERMISSION_DENIED
	details = "Permission monitoring.metricDescriptors.create denied (or the resource may not exist)."
google.api_core.exceptions.PermissionDenied: 403 Permission monitoring.metricDescriptors.create denied (or the resource may not exist).
```

Confirmed via direct Cloud Monitoring API query (`timeSeries.list` on `custom.googleapis.com/greenearth-api/feed.render.duration_ms`) that **existing** metrics/label combinations continue to receive fresh data points — this is not currently blocking the dashboards. However:
- Any **new** metric name or new label/attribute value introduced in future code will silently never get a descriptor created, and its data will be dropped with no other indication.
- The failure fires and logs a full multi-line Python traceback every single export interval in prod (every 60s, all day), which is significant, persistent log noise.

## Fix

1. Identify the Cloud Run service's runtime service account:
   `gcloud run services describe greenearth-api-prod --region=us-east1 --project=greenearth-471522 --format='value(spec.template.spec.serviceAccountName)'`
2. Grant it `roles/monitoring.metricWriter` (includes both `monitoring.metricDescriptors.create` and `monitoring.timeSeries.create`) on project `greenearth-471522`:
   `gcloud projects add-iam-policy-binding greenearth-471522 --member=serviceAccount:<SA> --role=roles/monitoring.metricWriter`
3. Add this grant to `scripts/gcp_setup.sh` (alongside the other IAM bindings it already manages, e.g. `ensure_inference_api_key_secret_access`) so it's reproducible rather than a manual one-off, and apply it for both `stage` and `prod` service accounts.

## Verification

After granting, confirm over a subsequent export cycle that the `PermissionDenied` / `metricDescriptors.create` traceback no longer appears:

```
gcloud logging read --project=greenearth-471522 --freshness=10m \
[...]

PR #200, opened 2026-06-30
add per-generator timeout, failure metrics, and gcp monitoring grant

### LINKED ISSUE #197: Add per-generator timeout/cancellation to run_generate so one slow generator can't stall the whole feed request
## Problem

`run_generate()` in `src/app/lib/candidates/generate.py` is called from the feed-skeleton pipeline (`_run_ranking_pipeline` in `src/app/routers/xrpc.py`) with `swallow_errors=True`. Inside `_run_one()` (generate.py:121-137), a failing generator's exception is caught, logged, and converted to `None` — but `asyncio.gather()` (generate.py:139-141) still waits for **every** generator task to fully finish before returning. There is no per-generator timeout or cancellation.

Confirmed in production logs from the #188 investigation (2026-06-29 23:31:43–23:32:11 UTC): a single `post_similarity` generator's Elasticsearch kNN query hung for 48–60+ seconds (`elastic_transport.ConnectionTimeout`), and that alone stalled the **entire** `getFeedSkeleton` response for the same duration. Twice this exceeded Cloud Run's 60s request timeout and returned a hard 504 to the Bluesky AppView (surfaced to the client as "upstream service unreachable"). Swallowing the generator's eventual error doesn't help when the slow call itself is unbounded — by the time it fails, the response budget is already gone.

A second, related gap: today, swallowed generator failures (`swallow_errors=True`) are only visible via `logger.exception(...)` lines — there's no metric, so nobody can tell from a dashboard how often generators are silently failing/timing out in production, or which ones.

## Fix

In `src/app/lib/candidates/generate.py`:

1. Wrap the `gen.generate(...)` call inside `_run_one()` (lines ~124-132) with `asyncio.wait_for(..., timeout=<per_generator_timeout>)`. Catch `asyncio.TimeoutError` alongside the existing `except Exception` handling: log a clear, distinct message (e.g. `logger.warning("Candidate generator '%s' timed out after %.1fs", spec.name, timeout)` — distinct from the generic `logger.exception(...)` used for other failures, so future incidents are immediately attributable instead of requiring traceback reconstruction), then return `None` when `swallow_errors=True` or raise `GeneratorError` otherwise, matching the existing behavior for other exception types.
2. Apply the identical timeout/cancellation wrapping to the infill generator call (lines ~162-176), which has the same unbounded-wait problem.
3. Make the timeout configurable via a new env var following the existing `GE_*` convention (e.g. `GE_CANDIDATE_GENERATOR_TIMEOUT_SEC`), defaulting to a value meaningfully shorter than Cloud Run's 60s request timeout (e.g. 10s) so one slow generator can't consume the whole request budget. Read it the same way other env-configured values in this codebase are read (see `_slow_threshold_ms()` in `src/app/lib/es_client.py` for the pattern).

Note: `asyncio.wait_for` cancels the wrapped coroutine on timeout, but the underlying `AsyncElasticsearch` HTTP call may not immediately release server-side resources — that's acceptable here since the goal is bounding *our* request latency, not cancelling the ES-side query.

## Metrics

Use the existing `MetricCollector` (see `get_metric_collector()` / `.record(name, value, **attributes)` usage in `src/app/lib/rankers/predict.py:156-161`) to make timeouts and swallowed failures countable and dashboard-able, not just visible in logs:

1. In `_run_one()`'s exception handling, emit a counter (name ending in `_count` per `MetricCollector.record`'s naming convention, e.g. `candidates.generator_failure_count`) on every swallowed-or-raised failure, with attributes `generator_name=spec.name` and `outcome` set to `"timeout"` (the new `asyncio.TimeoutError` branch) or `"error"` (the existing generic-exception branch), plus `is_infill="true"/"false"` so the infill path (lines ~162-176) reports through the same metric.
2. Guard with `if (mc := get_metric_collector()) is not None:` before recording, matching the existing pattern in `predict.py`.
3. This should let a dashboard answer "how often did generator X time out / fail in the last N minutes" without grepping logs — directly useful for confirming whether #199's ES retry/timeout tuning and greenearth-social/ingex#400's capacity work are actually reducing the swallowed-failure rate over time.

## Tests

Add/update tests in `src/app/lib/candidates/generate_test.py` covering:
- A generator that exceeds the timeout is cancelled and treated as a swallowed failure (returns `None`, logs the timeout warning, **and records the failure metric with `outcome="timeout"`**) when `swallow_errors=True`.
- The same case raises `GeneratorError` promptly (without waiting for some longer real timeout) when `swallow_errors=False` — confirm the metric is still recorded before the raise.
- A generator that raises a normal exception records the metric with `outcome="error"`.
- The infill generator path has equivalent coverage, including `is_infill="true"`.

mjmor:
Closes #197

Issue #196 (`monitoring.metricWriter` IAM grant) was already resolved at the code level in commit b3361ee — `gcp_setup.sh` already grants `roles/monitoring.metricWriter` to the API service account when `create_service_account()` runs. The remaining step is operational: re-run `./scripts/gcp_setup.sh --environment prod` to apply the grant to the existing prod SA.

The main code change here is #197.

# This PR

- **`src/app/lib/candidates/generate.py`**: Wraps each `gen.generate(...)` call in `_run_one()` with `asyncio.wait_for(..., timeout=_generator_timeout_sec())`. A distinct `WARNING` log fires on timeout (not `logger.exception`), making incidents immediately attributable. Adds equivalent timeout wrapping to the infill generator path. On both timeout and generic-exception paths, emits `candidates.generator_failure_count` (counter) via `MetricCollector` with attributes `generator_name`, `outcome` (`"timeout"` or `"error"`), and `is_infill` (`"true"/"false"`), so dashboards can answer "how often did generator X time out / fail" without grepping logs.
- **`GE_CANDIDATE_GENERATOR_TIMEOUT_SEC`** env var (default `10`s) controls the per-generator budget, well under Cloud Run's 60s request timeout.
- **`src/app/lib/candidates/generate_test.py`** (new): 9 tests covering timeout+swallow, timeout+no-swallow (raises `GeneratorError` promptly, metric recorded before raise), exception→error outcome, and all four cases for the infill path with `is_infill="true"`.

# Testing

- All 9 new tests in `generate_test.py` pass
- Full suite: 492 passed, 0 failures


src/app/lib/candidates/generate.py
```diff
@@ -18,11 +19,20 @@
...
 from .base import CandidateGenerator, CandidateResult, get_generator
 from ..feed_debug import current_recorder
+from ..metrics import get_metric_collector
 from ..telemetry import timed
 
 logger = logging.getLogger(__name__)
 
 
+def _generator_timeout_sec() -> float:
+    """Read per-call so the value tracks env changes during dev."""
```
mjmor: We don't need to read per-call at this time because the environment variable will not change until the redeployment. Please read once on startup and cache the value.

src/app/main.py
```diff
@@ -231,6 +231,14 @@ def _resolve_endpoint(request: Request) -> str | None:
...
         match, _ = route.matches(request.scope)
         if match is Match.FULL:
+            # FastAPI 0.137+ wraps included routers in _IncludedRouter which
+            # carries no name itself. Iterate its effective sub-routes to find
+            # the concrete one that matches.
+            if hasattr(route, "effective_route_contexts"):
+                for ctx in route.effective_route_contexts():
+                    orig = getattr(ctx, "original_route", None)
+                    if orig is not None and orig.matches(request.scope)[0] is Match.FULL:
+                        return getattr(orig, "name", None)
```
mjmor: This seems hacky and brittle. You're sure there's not a better way to match the current request to a route in the router?

mjmor (commented):
See comments below.

src/app/lib/candidates/generate.py
```diff
@@ -18,11 +19,20 @@
...
 from ..feed_debug import current_recorder
+from ..metrics import get_metric_collector
 from ..telemetry import timed
 
 logger = logging.getLogger(__name__)
 
 
+try:
+    _GENERATOR_TIMEOUT_SEC: float = float(
+        os.environ.get("GE_CANDIDATE_GENERATOR_TIMEOUT_SEC", "10")
```
mjmor: We need to set this as an environment variable in deploy.sh, right? Let's use 5 seconds as the max generator timeout too

src/app/lib/candidates/generate.py
```diff
@@ -123,15 +133,43 @@ async def _run_one(
...
+                return await asyncio.wait_for(
+                    gen.generate(
+                        es=es,
+                        user_did=request.user_did,
+                        num_candidates=count,
+                        video_only=request.video_only,
+                        exclude_uris=request.exclude_uris or None,
+                    ),
+                    timeout=_GENERATOR_TIMEOUT_SEC,
+                )
```
mjmor, line 145: Please also emit a metric for the candidate's generator success count. That way, we can compute an overall failure rate. Do the same for the infill generators as well.

mjmor: Also, let's use the same metric name structure as the duration metrics (`candidates.generate.duration_ms`) so for example the failure metric shoul dbe `candidates.generate.failure_count`

PR #216, opened 2026-07-03
Fail fast on pipeline errors behind GE_FAIL_FAST flag

### LINKED ISSUE #204: Add a flag for whether to serve a degraded feed vs failing
For now, we do not want to serve degraded feeds. If any component fails, fail rendering entirely and show the user an error.

We will want to turn this off eventually though. So make it configurable.

inthree3:
> **TL;DR**: Fail fast on feed pipeline errors, configurable via GE_FAIL_FAST env var (enabled in Cloud Run deployments).

## Fail-fast scope

**Covered (returns 500 when GE_FAIL_FAST=true)**
- Candidate generators — `generate()` failures (ES timeout, network error, etc.)
- Embedding hydration — `fetch_post_embeddings` ES call failure
- Ranker (`run_predict`) — always raises regardless of flag
- Perspective API — connection/timeout errors and non-429 HTTP errors

**Not covered (always soft-fail)**
- Perspective API 429 (rate limit) and language-not-supported — expected conditions, score treated as missing
- `_seen_exclusions` — Firestore read failure → proceeds with empty list (previously seen posts may reappear)
- `get_user` — debug flag read failure → proceeds with debug disabled
- Background tasks (`_record_session`, `_write_feed_debug`, `_record_interactions`) — fire-and-forget, no effect on response
- Feed cache — cache miss is a normal path; append/store failure returns results without a cursor
- Firestore / ID resolver initialisation checks — always 500 regardless of flag

## Toggling fail-fast

**To disable** (revert to degraded-feed serving): set `GE_FAIL_FAST=false` in `deploy.sh` and redeploy. No code change required.

| `GE_FAIL_FAST` | Behaviour |
|---|---|
| `true` | Any component failure → HTTP 500, no feed served |
| `false` (default) | Failures swallowed; feed served with whatever candidates survived |

The flag defaults to `false` so environments without the variable retain the previous behaviour.

## Implementation note

`fail_fast()` lives in `lib/config.py` so any module can import it directly without threading it as a parameter through the call stack.

Closes #204

### DIFF
--- a/src/app/routers/xrpc.py
+++ b/src/app/routers/xrpc.py
@@ -59,6 +59,14 @@ router = APIRouter(tags=["xrpc"])
 # Configuration
 # ---------------------------------------------------------------------------
 
+def _fail_fast() -> bool:
+    """When True, any pipeline component failure raises instead of being swallowed.
+
+    Controlled by the GE_FAIL_FAST env var (default: false).
+    Set to "true" to disable degraded-feed serving and fail on any component error.
+    """
+    return os.environ.get("GE_FAIL_FAST", "false").lower() in ("true", "1", "yes")
+
 
 def _get_service_did() -> str:
     """Return the DID of this feed generator service.
@@ -222,10 +230,9 @@ async def _hydrate_embeddings(es, candidates: list[CandidatePost]) -> list[Candi
         async with timed(logger, "hydrate_embeddings", n_missing=len(missing)):
             pairs = await fetch_post_embeddings(es, missing, index="posts_recent")
     except Exception:
-        # If the refetch fails, MMR falls back to author-only similarity
-        # and the two-tower ranker has its own refetch path. Don't fail
-        # the request over a hydration hiccup.
-        logger.exception("Embedding hydration failed; continuing without")
+        logger.exception("Embedding hydration failed")
+        if _fail_fast():
+            raise
         return candidates
 
     encoded: dict[str, str] = {}
@@ -274,7 +281,7 @@ async def _run_ranking_pipeline(
             num_candidates=gen_request.num_candidates,
             n_generators=len(gen_request.generators),
         ):
-            result = await run_generate(gen_request, es, swallow_errors=True)
+            result = await run_generate(gen_request, es, swallow_errors=not _fail_fast())
         candidates = result.candidates
 
         if not candidates:
--- a/src/app/routers/xrpc.py
+++ b/src/app/routers/xrpc.py
@@ -273,7 +273,7 @@ async def _run_ranking_pipeline(
             num_candidates=gen_request.num_candidates,
             n_generators=len(gen_request.generators),
         ):
-            result = await run_generate(gen_request, es, swallow_errors=not fail_fast())
+            result = await run_generate(gen_request, es)
         candidates = result.candidates
 
         if not candidates:
--- a/src/app/routers/xrpc.py
+++ b/src/app/routers/xrpc.py
@@ -46,6 +46,7 @@ from ..lib.firestore import (
     write_feed_debug,
 )
 from ..lib.posthog_client import get_posthog_client, track_interaction, track_session
+from ..lib.config import fail_fast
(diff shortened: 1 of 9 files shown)


src/app/routers/xrpc.py
mjmor, line 1: @inthree3 , would you also confirm the following cause an exception that bubbles up to a fast fail:
- connection/timeout errors on elastic search
- connection/timeout errors on perspective API
- any critical errors encountered during diversification / re-re-ranking component of the pipeline

Ensuring that critical exceptions in the first two bubble up may lead us to re-think this implementation a bit to prevent threading through parameters through the stack. E.g., we might want to have a request context variable to capture 'fail fast' and if set, allow exceptions to propagate up instead of log+suppress. Then the xrpc route can be responsible for handling exceptions and surfacing the 500 (as you've already done here).

inthree3: Sorry for the late reply, Max 🥲
Perspective API errors were falling outside the fail-fast scope, so I've added handling there. Thanks for catching it!
I agree the parameter threading was getting messy, so I went with a shared helper function in `lib/config.py` that any module can call directly.

mjmor (commented):
Pretty much ready to go @inthree3 ! You could merge this if you want and address these in follow ups (the second doesn't even need to necessarily be addressed if you'd prefer to keep moving instead).

One gap I don't think is covered yet: `followed_users_search` ([followed_users.py#L42-L55](https://github.com/greenearth-social/api/blob/main/src/app/lib/candidates/followed_users.py#L42-L55)) and `network_likes_search` ([network_likes.py#L155-L167](https://github.com/greenearth-social/api/blob/main/src/app/lib/candidates/network_likes.py#L155-L167)) both catch `FollowedUsersLookupError` (Bluesky follows-API failure) *inside* the generator function and return `[]`. That's below the `run_generate`/`_run_one` boundary that `fail_fast()` controls, so with `GE_FAIL_FAST=true` a Bluesky follows-API outage would still silently degrade those two generators to empty results instead of surfacing a 500. Let's re-raise `FollowedUsersLookupError` when `fail_fast()` is set.

Non-blocking style note: `fail_fast()` is called two different ways in this PR — `generate.py` threads it as a `swallow_errors` bool passed in by the caller, while `perspective.py` and `xrpc.py`'s `_hydrate_embeddings` import and call `fail_fast()` directly inline. Not a big deal, but picking one idiom consistently would make it easier to audit everywhere the flag actually takes effect.

inthree3:
> Pretty much ready to go @inthree3 ! You could merge this if you want and address these in follow ups (the second doesn't even need to necessarily be addressed if you'd prefer to keep moving instead).
> 
> One gap I don't think is covered yet: `followed_users_search` ([followed_users.py#L42-L55](https://github.com/greenearth-social/api/blob/main/src/app/lib/candidates/followed_users.py#L42-L55)) and `network_likes_search` ([network_likes.py#L155-L167](https://github.com/greenearth-social/api/blob/main/src/app/lib/candidates/network_likes.py#L155-L167)) both catch `FollowedUsersLookupError` (Bluesky follows-API failure) _inside_ the generator function and return `[]`. That's below the `run_generate`/`_run_one` boundary that `fail_fast()` controls, so with `GE_FAIL_FAST=true` a Bluesky follows-API outage would still silently degrade those two generators to empty results instead of surfacing a 500. Let's re-raise `FollowedUsersLookupError` when `fail_fast()` is set.
> 
> Non-blocking style note: `fail_fast()` is called two different ways in this PR — `generate.py` threads it as a `swallow_errors` bool passed in by the caller, while `perspective.py` and `xrpc.py`'s `_hydrate_embeddings` import and call `fail_fast()` directly inline. Not a big deal, but picking one idiom consistently would make it easier to audit everywhere the flag actually takes effect.

Confirmed the first issue and got it fixed! Thanks again for the review.

Regarding the second issue, the redundant functions seemed to confuse Claude too. I ended up removing `swallow_errors` and having run_generate call `fail_fast()` directly.

mjmor (approved):

PR #218, opened 2026-07-06
Popularity Candidate Generator in Your Feed

### LINKED ISSUE #219: Add Popularity as First-Class Candidate Generator in Your Feed

davidfreifeld:
Closes #219 

Add the `popularity` candidate generator as one of the main generators in Your Feed. Now we have 0.35 two-tower, 0.35 followed-users, and 0.3 popularity. Keep popularity as the infill generator. Because of this, added in logic to keep us from selecting the same posts that we already have at the infill stage. Also now returns an empty CandidateResult instead of None when swallow_errors is True, so that the feed_debug tool can differentiate between infill and primary generators.

Decided to add the `popularity` candidate generator into our feed because we anecdotally liked the results in the feed when the two tower generator was timing out and the results were coming from the popularity generator. (See discussion on discord).

Testing:
- Unit tests
- Deployed to stage
- Deployed locally connected to prod ES and sanity-checked the results

### DIFF
--- a/src/app/feeds.py
+++ b/src/app/feeds.py
@@ -29,8 +29,9 @@ FEEDS: dict[str, FeedConfig] = {
         internal_display_name="e2 S",
         gen_request_template=CandidateGenerateRequest.model_construct(
             generators=[
-                GeneratorSpec(name="two_tower", weight=0.5),
-                GeneratorSpec(name="followed_users", weight=0.5),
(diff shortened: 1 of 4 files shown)


inthree3 (approved):
LGTM! For the record, could you leave a short comment in this PR explaining why we decided to add `popularity` as a first-class feature?

PR #245, opened 2026-07-12
Add PostHog analytics integration for user growth and behavioral metrics

### LINKED ISSUE #243: Create simple growth metrics/dashboard
Metrics/widgets we need

- \# users (all time)
- active users (daily, monthly)
- \# sessions (all time)

In addition to these straight forward user growth metrics, we'll likely also want to log/analyze user behavior metrics in the same tool we choose. Some user behavior metrics will come from the bsky client and others will come from a separate client app we own (see more in "context" below). Bsky behavior that are sent to us:

```
requestLess
requestMore
clickthroughItem
clickthroughAuthor
clickthroughReposter
clickthroughEmbed
interactionSeen
interactionLike
interactionRepost
interactionReply
interactionQuote
interactionShare
```

## Context

All of the data needed to construct these metrics is captured in our firebase DB, but we also need it stored in an analytics DB to create time series metrics, etc.

We currently use GCP Cloud Monitoring to log system metrics, but nothing for users.

Also, at present we only have our API server where we serve bluesky feeds, so there's no "client" for us to install a client side user behavior tracking. Any metrics we track will need to have a back-end SDK available where we can behavior metrics. In the future (as part of #187 ) we'll also have a complementary front-end / client app, but even then, a large majority of user traffic will occur only in the BlueSky client where we don't have the ability to collect data client-side. This nuance will likely influence which analytics DB+tools we choose to adopt for growth metrics.

## This Task

1. Identify the appropriate analytics DB+tool to send user metrics. We should choose the analytics DB+tool that balances the tradeoff of being the quickest to implement while also having user analytics supported as a first-class entity. E.g., GCP Cloud Monitoring would be fast to set up because we already log there, but AFAIK there are no preconfigured user analytics dashboards/metrics. On the other hand, adding Google Analytics (G4A) may have a bit more up front cost because we haven't used it before, but it supports user analytics out of the box.
2. Implement the metrics in the API server on the selected analytics DB+tool.
3. Write a script to send all the user/session data we have in Firebase to the selected analytics DB+tool to backfill for past weeks before we implemented #2.

mjmor:
Closes #243

We chose PostHog as our analytics platform over GCP Cloud Monitoring (not built for user analytics) and GA4 (no reliable historical import, enterprise-only backend SDK). PostHog offers a first-class Python server-side SDK, built-in DAU/MAU/funnel reporting, feature flags + experimentation on all plans, and a self-hosting option that keeps all user data in GCP.

# This PR

- **`src/app/lib/posthog_client.py`** — process-level PostHog singleton with `track_session` (emits `feed_loaded` per feed request, drives DAU/MAU) and `track_interaction` (emits all 12 Bluesky behavioral events)
- **`src/app/main.py`** — init/shutdown PostHog client in lifespan from `GE_POSTHOG_API_KEY`; silently disabled if key absent
- **`src/app/routers/xrpc.py`** — `track_session` called from `_record_session` background task; `track_interaction` called from `_record_interactions` background task; both wrapped in try/except so PostHog failures never surface to users
- **`scripts/backfill_posthog.py`** — replays all historical Firestore users, feed activity, and interactions to PostHog with original timestamps; supports `--dry-run` and `--limit`

# Testing

- 14 new unit tests (8 for posthog_client, 2 for xrpc integration, 4 for backfill script); 579 total passing
- PostHog integration is gated behind `GE_POSTHOG_API_KEY` — absent key = silent no-op; existing test suite unaffected without the key set
- To enable: set `GE_POSTHOG_API_KEY=<project key>` and optionally `GE_POSTHOG_HOST` (default: `https://us.i.posthog.com`)
- Backfill: `pipenv run python scripts/backfill_posthog.py --dry-run` to preview; remove `--dry-run` to send


mjmor:
Reviewers (@raindrift , @inthree3 , @Gautham-Raju ) were chosen mostly for awareness since they're people who modify/depend on this code. I'm going to test/merge this pretty quickly so we have this data tracked and reviewable, but if you notice something worth changing, please file a story and I'll come back and fix!

PR #246, opened 2026-07-13
Add transparency endpoint for hydrating posts and pipeline metadata

### LINKED ISSUE #223: Support for feed transparency view in api
We have a way to store debug data for a feed load, but it's pretty heavyweight and can't be enabled all the time for every user.

Let's also store the minimal dataset needed to determine how we picked a given result, and cache it along with the rest of the data that we use for cursoring. That way we can re-render an already-served feed and present this data.

This story also includes two endpoints that our frontend can query. One returns a list of feed loads (`GET feed_loads`) that we still have cached (these expire after ~15 mins iirc), and the other (`GET feed_load?feed_load_id=foo`) will return all of the data (post uris, ranking metadata) for that specific feed load.

The `feed_load` endpoint should also return the configuration. For now it can just read it from feeds.py, but eventually it will be configurable, and we'll be pulling it from firebase instead.

These endpoints need to be authenticated, of course.

The `feed_load` endpoint should return the complete set of items. No need for paging since it's just a list of IDs and a little metadata.

Gautham-Raju:
- A web-based API (`/api/feeds`) surfaces pipeline observability for the frontend.
- Every feed load writes lightweight pipeline metadata, and the detail endpoint
returns it merged with hydrated Bluesky post data (author, media, engagement)
from the public API.  
- This is available regardless of whether `debug_feeds` is
enabled for the account.
- Responses use camelCase — the frontend can consume them directly without mapping.
- Authentication uses Firebase custom tokens (`Authorization: Bearer <token>`)
rather than API keys.  Token `uid` must be a `did:plc:…` — the prefix is
stripped to form the Firestore document key.


src/app/models_feed_debug.py
```diff
@@ -0,0 +1,111 @@
+"""Pydantic response models for the feed-debug transparency API.
+
+All models use camelCase JSON serialisation via ``CamelModel`` so the API
```
raindrift: I'm not sure that the inconsistency in the format of the API is worth the small reduction in typescript code complexity.

If it's important, I think I'd prefer that we do it as a middleware or something, where the api consumer can specify a camelcase flag that works with any api endpoint.

AGENTS.md
```diff
@@ -0,0 +1,188 @@
+# Green Earth API — Agent Notes
```
raindrift: Do we want to commit an AGENTS.md? I think there's a definite advantage to doing this, but I kinda want it to be in a different PR and for us to talk about what goes in there a little.

README.md
```diff
@@ -41,7 +41,13 @@ Start the development server with auto-reload:
 pipenv run uvicorn src.app.main:app --reload
 ```
 
-The API will be available at `http://localhost:8000`
+The API will be available at `http://localhost:8000`.
+
+> **Note**: The server refuses to start without `GE_ELASTICSEARCH_API_KEY` and
+> `GE_FEED_CONTEXT_SECRET` environment variables. For local development set
```
raindrift: FWIW, most local development does require an elasticsearch api key. I guess the stuff that only touches firebase doesn't, but everything else in the api queries ES, which is why this is required.

src/app/documents.py
```diff
@@ -269,3 +276,68 @@ class FeedDebugDocument(BaseModel):
...
+    at_uri: str
+    rank: int | None = None
+    rank_score: float | None = None
+    after_rank_position: int | None = None
+    generators: list[GeneratorMeta] = Field(default_factory=list)
+    model_scores: list[ModelScoreMeta] = Field(default_factory=list)
+    diversification: DiversificationMeta | None = None
+
+
+class FeedSnapshotDocument(BaseModel):
```
raindrift, line 362: I think it would make sense to expand or rework FeedCache for this purpose, since we're already writing it. Like, having two separate copies of the feed we are currently serving for slightly different purposes (cursoring vs transparency) doesn't seem like the right thing. I think it would be better to have a single copy of each served feed that handles everything.

Probably it makes more sense to keep what you have here though since it does a lot more, and refactor the FeedCache dependencies to make use of it instead.

This could happen in this PR, or you could make a story to come clean it up later. Obviously it's better to do it now though, since people have a habit of never actually doing those kinds of cleanup tasks. :)

src/app/documents.py
```diff
@@ -269,3 +276,68 @@ class FeedDebugDocument(BaseModel):
...
+    Separate from :class:`FeedDebugDocument` — that captures full
+    pipeline objects (content, author info, user features) for the CLI
+    debug tool and is only written for debug-flagged users.
+    """
+
+    request_id: str = Field(..., description="Feed-cache key / feedContext id (also the document ID)")
+    items: list[str] = Field(default_factory=list, description="AT URIs in final served order")
+    feed_name: str
+    generated_at: datetime
+    expires_at: datetime = Field(..., description="UTC expiration; drives native TTL")
```
raindrift, line 378: This is here, but I don't see a corresponding ensure_ttl_policy in gcp_setup.sh. Unless it's in a different PR. Otherwise nothing is going to actually expire these records.

src/app/routers/feed_debug.py
```diff
@@ -0,0 +1,211 @@
+"""Feed-debug transparency API endpoints.
```
raindrift: Probably want to call this file by a different name, since we have the feed_debug utility already that does something different. Maybe feed_transparency.py or similar?

src/app/lib/firebase_auth.py
```diff
@@ -0,0 +1,92 @@
+"""Firebase Auth token verification for the feed-debug transparency API.
```
raindrift, line 1: Nice! This is great! Seems very straightforward to use.

raindrift (changes requested):
This is looking good! I think the only real blocking issue is the lack of a TTL policy, though I have a bunch of other comments that you could address if you want to.

Thanks for putting this together!

src/app/documents.py
Gautham-Raju: Added

raindrift (approved):
Awesome! Thank you for putting this together!

PR #253, opened 2026-07-15
Pass Like Count to Heavy Ranker Inference Call from API

### LINKED ISSUE #222: Include like count in heavy ranker call
Once we've done the work to support like count as a feature in inference service [here](https://github.com/greenearth-social/inference-service/issues/20), modify the heavy ranker in the API to query for like count (already present on posts in ES) and pass it as an input in the inference call.

davidfreifeld:
Closes #222 

## This PR
- Queries like_count for the heavy ranker and passes it to inference service in the heavy ranker inference call
- There are two fetch_embeddings functions - one that just gets the embeddings, and one that gets the embeddings plus the like count and author. Same with the fetch_liked_post functions - one that just gets the likes and one that also gets the liked-at times. The two tower and heavy ranker models now both rely on the versions of the functions with extra fields. This means that when two tower is used as a candidate generator and heavy ranker as a ranker - we can cache and reuse the results from both ES queries instead of doing each of them twice.
- All candidate generators retrieve (and return) like_count now as well.

## Testing
- Generated candidates and ranked them locally
- Deployed to stage
- In both above scenarios - confirmed it works with inference service with the like count enabled or without it


mjmor (commented):
@davidfreifeld do you have a sense of how many ES queries really need all of the fields? Unless it's like 95% of calls need all the fields, I'd recommend approaching this differently (mostly because I don't want to ask more of ES performance right now).

If the goal is to reduce code, we could have a function that accepts kwargs identifying which fields to fetch for a given invocation and an object returned with nullable fields for optional fields.

PR #255, opened 2026-07-15
Rename Your Feed to GreenEarth

### LINKED ISSUE #254: Rename "Your Feed" to just "GreenEarth"
It's going to be a lot easier to communicate this way. Change it in the code, also the documentation.

inthree3:
Renames the `your-feed` display name from "Your Feed" to "GreenEarth" across feeds config and icons. The AT Protocol rkey (`your-feed`) is unchanged to avoid breaking existing subscribers.

**Changes**
- `display_name`: "Your Feed" → "GreenEarth"
- `display_name`: "Unranked YF" → "Unranked YF" (kept unchanged for better clarity)
- `internal_display_name`: "a0 YF" → "a0 YF" (kept unchanged for the same reason)
- Icon: `your-feed.png` → `green-earth.png`

**Results**
image

Now, the feed is online as 'GreenEarth' 🌍

Closes #254


src/app/feeds.py
```diff
@@ -24,7 +24,7 @@
 FEEDS: dict[str, FeedConfig] = {
     "unranked-your-feed": FeedConfig(
         display_name="Unranked YF",
```
raindrift, line 26: Probably want to change YF to GE here, for consistency.

src/app/feeds.py
```diff
@@ -59,12 +59,12 @@
         ),
     ),
     "your-feed": FeedConfig(
```
raindrift, line 61: Do we want to change the key here too? I think maybe no, since it would make the feed disappear for all of its current subscribers. OTOH, if we ever want to change it, now is a better time than at any future point.

src/app/feeds.py
inthree3: I see. Initially, I changed it to GE as per your comment, but it seemed a bit confusing since GE could also be read as a family name reference, so I reverted it and only changed the display name.

However, I agree that consistency is important, so I'll go ahead and change it back to GE.

src/app/feeds.py
inthree3: I discussed with Jonathan about this issue, and we agreed not to change it!

src/app/feeds.py
raindrift: Yeah, you have a good point. I think, do whatever you think is best. It's only for the team anyway, and it's easy to change it later.

inthree3: Okay, then let's leave it and we can fix it if there's any disturbance.

raindrift (approved):

Issue #257, opened 2026-07-15
Change ranker scoring range to 0..1 so it makes sense in the UI


raindrift:
Are scores -1 to 1? 0-1? Do the scales for them match so they look comparable when users see them? This is also a good check for whether we're doing our own math right.

This story is to look at them, then adjust them so they make sense.

For example, we normalize heavy ranker candidates based on min/max for that batch, rather than calibrating to some kind of external constant. How widely does that swing? Is that good or bad?

Gautham-Raju:
Relatedly, the engaging and constructive scores can be negative. Should we normalize this on the frontend because of the slider UI displaying it? Or should we let the backend handle this?

PR #265, opened 2026-07-17
Configurable limits on which results are returned

### LINKED ISSUE #248: Do not render all candidate posts, define minimum scores
We should not render all candidate posts, since we can assume that some of the candidates will be bad for one reason or another. We should probably return somewhere between 10-50% of them.

We should define a set of minimums, below which we stop returning results and instead restart the ranking session (generate a new slate). Minimums we want to define:

- Share (as a percentage) of the retrieved candidates to return. Like if this is set to 0.2, we never return more than 20% of the candidates we were able to retrieve.
- Minimum MMR score. If we start seeing candidates below this score, restart the ranking session
- Minimum combined ranking score. If we start seeing rankings that are just plain bad, restart the session

These should be configurable in feeds.py.

We should log a metric when we hit any of these limits. 

Right now, we store the ranker output in firebase, and then page through it with the cursor. We also keep track of all posts the user has actually seen, and exclude those from subsequent candidate queries.

When we do this, something we will have to be concerned with is that the posts we get on subsequent queries might be very similar to the ones we had before. Make sure we have metrics in place to determine if we're being starved for candidates, or doing a lot of knn search retries. Some of our existing metrics may be enough for this, unclear.

When running queries for subsequent ranking sessions, exclude posts that had a low ranker score (we will never display them, so no point fetching or ranking them again). We may want to store these alongside the seen posts, to make future queries faster. If we do that, consider creating an after_request hook for these bookkeeping queries, so we don't slow down the feed with them.

It's ok to make changes around the seen_posts subcollection, but if possible let's avoid making changes in the cache used for cursoring, since #223  (and its PR, #246) may end up reworking that code significantly.

raindrift:
Closes #248.

## Motivation

The feed pipeline previously rendered every retrieved candidate: the whole ranked batch was cached and the cursor paged through it to the end, including the low-quality tail. This PR cuts slates off at configurable quality gates, restarts the ranking session when a slate runs short, and stops re-fetching posts we already know we'll never display.

## What's here

**Per-feed quality gates** (configurable in \`feeds.py\`, \`None\` disables):
- \`max_render_share\` — cap on the share of retrieved candidates rendered per slate
- \`min_rank_score\` — combined rank-score floor; sub-floor candidates are cut *and recorded as discarded*
- \`min_mmr_score\` — the slate stops at the first MMR pick below this (slate-relative; MMR relevance is normalized per slate)

Enabled on \`your-feed\` and \`best-of-friends\` at \`0.5 / 0.0 / -0.05\` — starting points to tune from the new metrics. If the gates reject everything retrieved, we fail open and serve the pre-cutoff slate (module constant \`EMPTY_SLATE_FAIL_OPEN\` flips this to an empty feed).

**Session restart via existing pagination.** Short slates now still store to the feed cache and issue a cursor, so paging past the end lands in the existing regeneration branch with cached + seen + discarded posts excluded — no \`lib/feed_cache.py\` changes (avoids conflicting with #223/#246). Fixes in passing: a pre-existing pinned-post off-by-one where page 1 consumed \`limit-1\` generated posts but the cursor advanced by \`limit\`, silently skipping one post on page 2.

**Discarded-posts store.** Posts below the rank floor are written (off the hot path, fire-and-forget) to a new \`users/{id}/discarded_posts/{YYYY-MM-DD}\` subcollection mirroring seen_posts, with a 3-day TTL, and merged into \`exclude_uris\` on future generation. MMR-cut and share-capped posts are *not* discarded — they were relevance-worthy and may fit a later slate.

**Metrics** (starvation + gate visibility): \`feed.slate.cutoff_count\` (\`feed_name\`, \`reason\`), \`feed.slate.kept_share\`, \`feed.slate.empty_after_cutoff_count\`, \`candidates.generate.retrieved_share\`, \`feed.slate.exclusion_size\`.

**Debug visibility.** \`mmr_rerank\` now returns \`(candidate, pick_score)\` pairs; the debug recorder captures per-reason cutoff URIs and \`n_retrieved\`, and \`scripts/feed_debug.py --show\` renders a \`cutoffs\` header line (\`rank_score=5  mmr_score=6  kept 23/50 (46%)\`) plus a \`reason\` column in the discarded table.

## Ops note

\`gcp_setup.sh\` gains \`ensure_ttl_policy discarded_posts\` — **must be applied in stage and prod**, or buckets accumulate (degrades to stale exclusions, not breakage; reads cap at 1000).

## Verification

Full test suite passes (682). Also driven live against prod ES (port-forward) + stage inference + real Perspective API with the Firestore emulator: a 50-candidate request served 15 posts (5 cut by rank floor, 6 by MMR floor), wrote 18 discards with the 3-day TTL, and subsequent requests never re-retrieved them; six pages of scrolling restarted the session twice with zero duplicate posts; pinned page 2 continued exactly where page 1 left off.

## Notes for reviewers

- Cost: at share 0.5 the full pipeline runs every ~50 posts instead of ~100; watch \`feed.render.duration_ms\` after rollout.
- \`min_rank_score=0.0\` is the theoretical midpoint of the [-1, 1] combined score, not an empirical one — tune from \`feed.slate.kept_share\` once live.
- Probe requests accumulate discards for the probe DID; harmless within the 3-day window.


src/app/lib/diversify.py
```diff
@@ -16,9 +16,18 @@
 DECAY_TAU = 15.0
 
 
-def mmr_rerank(candidates: list[CandidatePost]) -> list[CandidatePost]:
+def mmr_rerank(candidates: list[CandidatePost]) -> list[tuple[CandidatePost, float]]:
```
mjmor, line 19: We might want to wait for #246 to get merged in, then this plumbing will become a lot easier. I needed the diversity score to bubble up in #244 as well and ended up depending on #246 and bubbling it up there.

src/app/models.py
```diff
@@ -218,6 +218,27 @@ class FeedConfig(BaseModel):
         None,
         description="AT URI of a post to pin at the top of the first page of this feed.",
     )
+    max_render_share: float | None = Field(
+        None,
+        gt=0.0,
+        le=1.0,
+        description="Cap on the share of retrieved candidates that may be rendered "
+        "(e.g. 0.5 → at most 50% of the retrieved candidates are returned per slate). "
```
mjmor, line 226: Should we have a global ratio cap on candidates? Shouldn't the minimum rank and diversity scores suffice to accomplish the same / be the right mechanism to achieve our goal of filtering out crummy end-of-batch content?

mjmor (commented):
The first comment is just a code hygiene / code conflict suggestion, but curious to hear your thoughts on the `max_render_share` comment. At the very least, we could leave the code as is and set max render share to 100%. That'd let the rank and MMR thresholds do all the heavy lifting, then we could review `cutoff_count` / `kept_share` and decide whether `max_render_share` is needed.

src/app/models.py
raindrift: I suspect that eventually, probably, and then we'll be able to remove this. But for the moment we're not sure where to set the score-based ones, so I wanted this other big hammer just in case.

src/app/lib/diversify.py
raindrift: Good point. I'll stack this one #246 instead.

davidfreifeld:
@raindrift FYI as per Jonathan and Gautham's request, I am working on updating the ranker scores to be in 0..1 instead of -1..1 (See #257). In case that impacts this PR

raindrift:
> @raindrift FYI as per Jonathan and Gautham's request, I am working on updating the ranker scores to be in 0..1 instead of -1..1 (See #257). In case that impacts this PR

Good to know! I think it will be fine to merge the PR for now, but we'll want to update the thresholds once that work is done.

Issue #144, opened 2026-06-05
Investigate (and maybe correct) low author diversity in Your Feed


raindrift:
Jonathan was seeing a lot of new posts from the same outlet. Check to see if this is a candidate gen or diversification issue. If it's diversification, fix it. If it's candidate gen, it's a known issue and we can close this task.

Note: See #148. It's possible that the post_similarity candidate generator is down, which means the Your Feed candidate pool is purely followed_users, with popularity infill. So fixing #148 might alleviate or resolve this issue.

raindrift:
Plz use the new feed debug tools to find out where this issue is occurring in the pipeline for you.

davidfreifeld:
Posted on discord and in our dogfooding notes but some thoughts on diversification with respect to authors: 

It currently penalizes a post if the author of that post already exists in the results. But it is binary - it doesn’t matter if the author is in the results 1 or 100 times. I think we should penalize depending on how many times that author has already shown up in the results. We might also need to increase the penalty in general, it's very often not enough to keep posts from making it into the feed. 

Note: we might need to increase the generated candidates count too if the diversification becomes more strict. Otherwise the similar posts will show up at the end of the list anyway.

raindrift:
I suspect this is an issue with the generated candidate count being too small. Also I think we run through every candidate before refreshing. We should instead serve the top-k candidates and then reload. Otherwise we will inevitably end up with the dregs, which would cause exactly this issue.

PR #286, opened 2026-07-22
Log mean diversity score per feed batch

### LINKED ISSUE #206: Log the mean diversity score for each batch of returned results
Each time we return a set of results to the client (endpoint is cursored, so for each batch), compute the mean diversity score and log it as a metric. Also maybe tag with the cursor position (how many batches were returned). Then we can see if we are returning insufficiently diverse results.

This is to help correct #144 

If the solution for #144 turns out to be obvious, this may not be needed and we can icebox it or delete it.

mjmor:
Closes #206

The feed generator pre-computes MMR diversity scores and serves results in cursor-paginated batches. Without observability into how diverse each served page is, it's hard to detect whether returned results are insufficiently diverse (the motivation behind #144).

# This PR

- Extend the diag tuple in `mmr_rerank()` (now returning `(candidate, pick_score)` pairs per the configurable-limits work already on `main`) to also compute and record `diversity_score` per pick
- Thread `diversity_score` through `FeedDebugDiversificationEntry` and `DiversificationMeta`/`PipelineItemMeta`, so it rides along with the rest of the per-item pipeline metadata already captured for every feed load
- Emit `feed.mean_diversity_score` histogram at all three serve points in `getFeedSkeleton`, reading scores directly from whichever `items_meta` is already in scope (`cache_doc.items_meta` for cache-hit pages, `generated_snapshot.items_meta` for freshly-generated pages) — no extra Firestore read needed, since `FeedCacheDocument` already carries `items_meta`
- Pinned posts are excluded from the mean via an explicit `exclude_uri` parameter, verified even when the pinned post happens to also be an organically-generated candidate
- Removed the now-dead `CandidatePost.diversity_score` field (superseded by the diag-tuple/items_meta approach; nothing produces it anymore since `mmr_rerank` stopped stamping `CandidatePost` objects)

# Testing

- Unit tests for `mmr_rerank` diversity score computation (`src/app/lib/diversify_test.py`)
- `feed_debug_test.py`: diversity_score flows correctly into both `FeedDebugDiversificationEntry` and `DiversificationMeta`
- Metric emission tests: initial request, cursor page served from cache `items_meta` (no extra I/O), diversification-disabled (no metric), and pinned-post exclusion verified against ground-truth scores computed via the same production `mmr_rerank` call — `src/app/routers/xrpc_test.py::TestDiversityScoreMetric`
- Full suite: 670 passed, pyright clean


src/app/lib/diversify.py
```diff
@@ -84,11 +86,18 @@ def _calculate_penalized_score(i: int) -> float:
...
             content_penalty = 0.0
+            diversity_score = 1.0
         else:
             best = max(remaining, key=_calculate_penalized_score)
             pick_score = _calculate_penalized_score(best)
             author_penalty = _calculate_author_penalty(best)
             content_penalty = _calculate_content_penalty(best)
+            # Diversity score mirrors the combined similarity the penalty was
+            # derived from, so it stays comparable to the pre-decay formula.
+            combined_sim = decayed_same_author_counts[best] * AUTHOR_WEIGHT + (
```
davidfreifeld: I think this is fine, but just pointing out that combined_sim can be greater than 1.0. For example, if we choose the same author 3 times in a row, the 3rd pick will have decayed_same_author_counts equal to 1.0 + 0.94. Then a modest content sim will put the total greater than 1.0. We floor diversity_score at 0.0. So we just lose a little information about the *really* homogenous slates.

Is there a strict reason that we need the score to be in 0..1? Or that we want a higher score to mean "better"? If not, we could possibly just use the combined_sim directly. But as I said I think it's also fine to leave it as-is.

src/app/routers/xrpc.py
```diff
@@ -641,6 +641,51 @@ def _get_feed_cache(request: Request) -> FeedCache:
...
+    joined) are simply absent from the result.
+    """
+    return {
+        meta.at_uri: meta.diversification.diversity_score
+        for meta in items_meta
+        if meta.diversification is not None
+    }
+
+
+def _record_diversity_metric(
```
davidfreifeld: Again I think this is fine but I want to point it out: we record one average data point for each page, regardless of page size. So pages of 30 posts or 2 posts get equal weight. And I'd expect the average diversity of the former to be worse on average - there are just more posts for each of the (especially low down) posts to be similar to. But this is a metric we'll be monitoring over time, right? So all of that should come out in the wash. But we should keep in mind that if our cursor/pagination behavior changes, that might effect these numbers too.

davidfreifeld:
LGTM! Pointed out two things but I think neither is a blocker from merging.

mjmor:
@davidfreifeld I agree with your ponts. Did you see any obvious/low lift ways to address those? I didn't offhand, but if there's something easy, I could add it with the PR before merge.

davidfreifeld:
I think for the first point, we could just log `combined_sim` instead of `max(0.0, 1.0 - combined_sim)`. Unless there is a reason the score needs to be in 0..1 or that it needs to be "higher is better". But we might want to rename it - `diversity_penalty` or `similarity_score` or something like that. Still thinking about the second comment.

davidfreifeld:
And to the second point, the best thing I can think of would be to come up with some page size buckets (e.g. 1, 2-5, 6-15, 16-30, 31+) and then in mc.record() add something like `score_count_bucket="16-30",`. Then I think we'd have the data necessary to see diversity split out by page size, right? 

But I don't feel so strongly, I think it's fine to leave out.

PR #299, opened 2026-07-23
Gate fail-fast behind PostHog feature flag (issue #279)

### LINKED ISSUE #279: Gate fail-fast behind a PostHog feature flag scoped to internal users
Closes #271 (team decision on fail-fast rollout)

Context: #216 added `GE_FAIL_FAST` as a server-side env var, but it was never deployed to prod because the underlying errors it would surface (Perspective API timeouts, intermittent ES connection errors) are still being swallowed softly (#272 defaulted it back to `false` as a safety measure). #271 noted that at ~60 DAU we can't afford to break every user's feed while shaking out failure modes.

The right next step isn't to enable or disable fail-fast globally — it's to enable it only for the team so we get real signal on the failure modes without exposing general users to broken feeds.

# This Issue

Replace the `GE_FAIL_FAST` env-var boolean with a PostHog feature flag, evaluated per-request against the authenticated user. The flag should be enabled for the **Internal Test Users** cohort (team members) in PostHog and disabled for everyone else.

- Add a PostHog feature flag (e.g. `fail-fast-feed`) configured to roll out to 100% of the Internal Test Users cohort and 0% of all other users
- In the feed request handler, resolve the authenticated user's DID/UID and evaluate the feature flag via the PostHog server-side SDK before invoking the pipeline
- Pass the resolved boolean into `lib/config.py`'s `fail_fast()` (or replace its env-var read with the per-request value) so the existing fail-fast code paths are unchanged
- `GE_FAIL_FAST` env var can be removed from `deploy.sh` once the feature flag controls the behavior

## Why PostHog cohort instead of env var

| Approach | Who sees fail-fast |
|---|---|
| `GE_FAIL_FAST=true` | Everyone |
| `GE_FAIL_FAST=false` | Nobody |
| PostHog flag + Internal cohort | Team only |

This lets us validate failure surface area against real prod data using team accounts, fix the underlying errors (#277, #278), and then graduate the flag to a broader rollout — all without a redeploy.

## Acceptance Criteria

- [ ] PostHog feature flag `fail-fast-feed` created and scoped to the Internal Test Users cohort
- [ ] Feed handler evaluates the flag per-request using the authenticated user's identity
- [ ] Fail-fast behavior is active for team accounts in prod and inactive for all other users
- [ ] `GE_FAIL_FAST` removed from `deploy.sh` (cleanup)
- [ ] Tested end-to-end with a team account against the stage or prod environment

inthree3:
## Summary

- Replaces the `GE_FAIL_FAST` env-var boolean with a per-request PostHog feature flag (`fail-fast-feed`), so fail-fast is active only for Internal Test Users, not general users
- Adds a `ContextVar` in `lib/config.py` to hold the per-request override; all existing `fail_fast()` call sites pick it up automatically with no changes
- Adds `evaluate_fail_fast_flag()` to `posthog_client.py`, soft-fails to `False` on any SDK error so a PostHog outage never breaks feed serving
- Removes `GE_FAIL_FAST` env var since it serves little now that the PostHog feature flag plays the same role.

Closes #279

## PostHog setup (manual, after merge)

1. Create feature flag `fail-fast-feed` in the PostHog project (stage + prod)
2. Set rollout: 100% for **Internal Test Users** cohort, 0% for everyone else
3. Enable the flag


src/app/lib/posthog_client.py
```diff
@@ -87,3 +90,19 @@ def track_interaction(
...
+    """Evaluate the fail-fast-feed PostHog feature flag for this user.
+
+    Returns True only when the client is present and the flag is enabled for
+    user_did. Soft-fails to False on any SDK exception so a PostHog outage
+    never breaks feed serving.
+    """
+    if client is None:
+        return False
+    try:
+        return bool(client.feature_enabled("fail-fast-feed", user_did))
```
mjmor, line 105: Just curious: is the feature flag on the posthog client also cached per request? If so, we could remove the extra logic setting the request level variable.

mjmor (approved):

src/app/lib/posthog_client.py
inthree3: Thank you for the good question, Max!
The `set_fail_fast_for_request` / `ContextVar` logic exists to propagate the flag value deep into the pipeline without threading `client` and `user_did` through every function signature. 

Calling `evaluate_fail_fast_flag()` directly at each call site would require passing those two arguments all the way down the stack.

PR #300, opened 2026-07-23
Bound feed pipeline stage latency with per-stage timeouts

### LINKED ISSUE #291: Investigate and fix "feed unavailable" error
## The Problem

When attempting to load https://bsky.app/profile/greenearth-social.bsky.social/feed/your-feed, the client sends the following request and receives a 502 response:

[bsky.app_xrpc_app.bsky.feed.getFeed-2026-07-22-19-37-03.har.zip](https://github.com/user-attachments/files/30287166/bsky.app_xrpc_app.bsky.feed.getFeed-2026-07-22-19-37-03.har.zip)

Response body: `{"error":"UpstreamFailure","message":"feed unavailable"}`

This happens 2-3 refreshes in a row until a successful feed load is achieved.

It's also clear that we're returning a list of posts for those same requests that present as "feed unavailable" in the client, because we can see the feed results recorded in the debug firestore collection when feed_debug.py tool is enabled. We can also see in the GCP logs that only 200 HTTP responses are returned (based on #270). This points to the likely fact that we're returning a feed skeleton and 200 response, but something after the getFeedSkeleton call does not like that response and throws an error to the client.

AFAIK, the only step between `getFeedSkeleton` (the endpoint we expose in the API service) and `/xrpc/app.bsky.feed.getFeed?feed=at://did:plc:wrmpulygwvuhjn2c3jbalgqj/app.bsky.feed.generator/your-feed&limit=30` (the endpoint called from the bsky client) is a post hydration step and probably some response validation.

## Investigation

Let's explore whether there's a place in our code where we're not returning a JSON object that does not conform with the bsky lexicon expected for [getFeedSkeleton](https://github.com/bluesky-social/atproto/blob/main/lexicons/app/bsky/feed/getFeedSkeleton.json) or if there's specific client implementation details that point to why our feed would fail to load on the client but succeed on the API server

### Findings

Traced the exact AppView code path that produces this error (`bluesky-social/atproto`, `packages/bsky/src/api/app/bsky/feed/getFeed.ts`): `"error":"UpstreamFailure","message":"feed unavailable"` is only emitted by a generic network/timeout failure calling our `getFeedSkeleton` endpoint — not by a lexicon/schema mismatch (that produces a different message, `"feed provided an invalid response"`), and not by the hydration/blocks-and-mutes step (which silently filters unhydratable items rather than failing the whole request). The AppView calls us with a [hard 10s abort (`AbortSignal.timeout(10_000)`) and zero retries](https://github.com/bluesky-social/atproto/blob/a667bc2040e0c7ff30499fb8f5628d90dd549c2d/packages/bsky/src/api/app/bsky/feed/getFeed.ts#L211-L221), and we've confirmed our `getFeedSkeleton` latency regularly exceeds 10s — our server still finishes the work and logs a 200, but by then the AppView has already aborted and returned the error to the client.

### Next steps

1. Update #287 to lower `GE_FEED_REQUEST_TIMEOUT_SEC` from 45s to 9s (below the AppView's 10s abort) so we fail fast with a logged, metered 504 instead of silently losing the race against the client's timeout.
2. Open a new PR to enforce ~3s sub-budgets on each pipeline stage (candidate generation, ranking, diversification), degrading gracefully (skip/shorten a stage) rather than exceeding the outer 9s budget.
3. Expect the "degraded feed" rate to rise significantly in the short term — an accepted tradeoff, since a degraded-but-served feed beats a client-side "feed unavailable" error.
4. Use the metrics from #287/#280 to identify which pipeline stage(s) are slow and drive follow-up perf work to bring feeds back under 10s undegraded.

mjmor:
Stacked on #280, which already added the `PipelineContext`/`DegradationEvent` scaffolding, the `feed.render.degraded_count` metric, and try/except-based fallbacks for the rank and embedding-hydration stages. This PR bounds each stage's wall-clock time so those fallbacks actually trigger before the outer 9s request timeout (#287) fires and drops the whole response.

Note: the budgets allowed to each stage were chosen based on some very basic analysis of data determining which stages take the longest. 

# This PR

- Lower `GE_CANDIDATE_GENERATOR_TIMEOUT_SEC` default from 15s to 4s in `generate.py` and `scripts/deploy.sh` — no new code path needed, the existing per-generator timeout/degradation handling in `generate.py` already covers this.
- Add a per-ranker-model timeout in `predict.py` (`GE_RANK_MODEL_TIMEOUT_SEC`, default 2.5s), wrapping each ranker's own coroutine rather than the whole `run_predict()` call. `asyncio.gather(..., return_exceptions=True)` means one ranker timing out doesn't cancel a concurrently-running, already-completed ranker's results — a timed-out model is simply excluded from the weighted combination, the same as a model that returns zero valid scores today.
- Add an embedding-hydration timeout in `xrpc.py`'s `_hydrate_embeddings` (`GE_EMBED_HYDRATION_TIMEOUT_SEC`, default 1.5s), wrapping the single batched `fetch_post_embeddings` call. The resulting `TimeoutError` falls into the existing `except Exception` handler, which already logs, records a `DegradationStage.EMBED_HYDRATION` event, and returns candidates unhydrated.
- No changes to diversification (MMR) — it's synchronous, in-process, and O(n²) over ~30 candidates; wrapping it in `asyncio.wait_for` wouldn't actually interrupt it mid-loop, so it's not a real timeout risk.
- No new metric — the existing `feed.render.degraded_count` plus the per-stage log lines from #280 (and the new per-ranker/hydration timeout log lines added here) are sufficient to distinguish degraded-vs-timed-out responses.

Budget check against the outer 9s timeout: 4s (gen) + 1.5s (hydrate) + 2.5s (rank) = 8s, leaving ~1s slack for MMR + cutoffs + serialization.

# Testing

- New unit test in `predict_test.py`: a hanging ranker model times out and is excluded from the combined score, while a concurrently-running fast ranker's results are preserved (not discarded).
- New unit test in `xrpc_test.py`: a hanging `fetch_post_embeddings` call times out and `_hydrate_embeddings` falls back to unhydrated candidates via the existing degradation path.
- Full suite: `pytest src/` — 686 passed.
- `ruff check` on all changed files — no new lint errors introduced (pre-existing lint debt in these files left untouched).


raindrift (approved):
Thanks for making these configurable. I think we might end up wanting them to be different in dev, so it'll be easy to do that. :)

Issue #301, opened 2026-07-23
Dev environment: include the frontend and use its Firebase emulator config


raindrift:
Part of the dev environment work (#268). The frontend repo is now substantially built out (React/Vite app, Firebase Functions, firestore rules, Playwright e2e), and per greenearth-social/frontend#42 the Firebase configuration is consolidating there — the api repo's copy is going away. The dev environment needs to follow both changes.

This supersedes the "Frontend compose profile" bullet in #284, which was written when the frontend repo was an empty stub.

**Depends on greenearth-social/frontend#42** (in progress), which moves firestore rules, indices, emulator config, and TTL policies out of the api repo.

## Why the emulator has to move

The devenv today runs a bare `gcloud emulators firestore start` — it uses no repo's Firebase config at all. That means it currently gets:

- no `firestore.rules`, so the api's auth-dependent paths aren't exercised against real rules
- no indices, so a query needing a composite index can pass locally and fail deployed
- no Auth emulator and no Functions emulator, which the frontend needs for the Bluesky OAuth flow (`authBluesky`, `oauthCallback`, `oauthClientMetadata`, `oauthJwks`)

The frontend's `firebase.json` already declares all three emulators (auth 9099, firestore 8080, functions 5001, `singleProjectMode`). Starting `firebase emulators:start` from the frontend checkout gets rules, indices, auth, and functions from the single source of truth that #42 is establishing, instead of us maintaining a fourth copy of the config inside devenv.

## Scope

- Replace the bare firestore emulator service with one that runs `firebase emulators:start` from the mounted frontend checkout. It needs the Firebase CLI and the functions' own dependencies installed, so this is a purpose-built image rather than the stock cloud-cli one.
- Add the frontend itself (Vite dev server) to the compose stack, wired to the local api, honoring the existing `--watch` convention.
- Promote `frontend` from optional to required in `devctl`'s repo checks, and extend `devctl nuke`/`status` to cover the new services.
- Support both frontend modes: `VITE_USE_MOCK_SERVICES=true` (the repo default — full UI, no backend) should still work for pure UI work, with the wired-up mode available for end-to-end work.
- Update devenv README: the Firestore section, the service/port table, and the api-repo emulator instructions that #42 makes obsolete.

## Integration details worth knowing up front

[...]

PR #302, opened 2026-07-23
Record a user session even when their handle won't resolve

raindrift:
Base branch: `main`.

`_record_session` resolved the caller's handle before writing anything and **returned early if that failed** — so one failed lookup skipped the user document, the feed-activity write, and the PostHog event entirely.

Handle resolution goes over the network to the PLC directory, so it fails for reasons that have nothing to do with the user existing: a directory blip, a DID document that's briefly unresolvable. Losing the handle shouldn't also lose the record of the visit, which is keyed on the DID we already have. `UserDocument.username` has always been `str | None`; only this path insisted on it.

Resolution failure is now a warning and the session is recorded with a null handle. Two places had to stop treating null as a value to write, so a transient failure can't **erase** a handle we already know:

- `upsert_user` leaves an existing username alone when passed `None` (it still refreshes `last_seen_at` — the visit did happen).
- `track_session` omits the `$set` person property rather than nulling it, while still capturing the `feedLoaded` event.

## How this surfaced

Found while bringing the frontend up in the shared dev environment ([#301](https://github.com/greenearth-social/api/issues/301)). Its seeded personas are pseudonymized, so they never resolve — which meant no user document was ever created, and `scripts/feed_debug.py --enable` had nothing to attach to. That made the failure mode easy to see, but it isn't dev-only: the same thing happens in production whenever the directory is briefly unavailable, silently dropping user records and feed activity for affected requests.

## Tests

792 pass. Added coverage for each guarantee: a user document is created with no handle, an unresolved handle doesn't overwrite a known one, `track_session` still fires without setting the person property, and `_record_session` completes all three writes when resolution raises.

One existing test asserted the old log wording; updated to match, and its docstring now states the stronger guarantee.

Verified in the dev environment: after the change the api creates the user document on its own (`username: null`) and `feed_debug.py --enable` succeeds with no manual setup.

### DIFF
--- a/src/app/lib/firestore.py
+++ b/src/app/lib/firestore.py
@@ -107,12 +107,17 @@ async def get_user(db: AsyncClient, user_did: str) -> UserDocument | None:
     return UserDocument.model_validate(data)
 
 
-async def upsert_user(db: AsyncClient, user_did: str, username: str) -> UserDocument:
+async def upsert_user(db: AsyncClient, user_did: str, username: str | None) -> UserDocument:
     """Create or update a user document.
 
     On first visit the document is created with all timestamps set to now.
     On subsequent visits ``last_seen_at`` is refreshed and ``username`` is
     updated if it changed.
+
+    ``username`` may be ``None`` when the caller's handle couldn't be resolved
+    (the DID is the identity; the handle is enrichment). A ``None`` never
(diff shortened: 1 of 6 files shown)


raindrift:
Merging this since I need it to move forward with devenv setup

PR #303, opened 2026-07-23
Add a development session path for getFeedSkeleton

raindrift:
Base branch: `main`. Needed by the dev environment ([#301](https://github.com/greenearth-social/api/issues/301) / greenearth-social/internal-tools#4).

`getFeedSkeleton` authenticates an AT Protocol JWT signed by the caller's Bluesky identity. Nothing outside production can mint one, so the **signed-in-user path — the one that records feed snapshots, user documents and activity — has been impossible to exercise locally.** The frontend's transparency UI reads those snapshots, so it has nothing to show in a dev environment.

`GE_DEV_SESSION_SECRET` lets the api accept an explicit stand-in: `X-Dev-Session` carrying the secret, `X-Dev-Session-DID` naming the user. The request is then treated as a genuine session everywhere downstream.

## Why not extend the probe bypass

It would have been less code, and it's tempting because the bypass already skips auth. But:

- **It means the opposite thing.** A probe is monitoring traffic, deliberately excluded from user data (`if not is_probe`). This needs to be *included* in user data — that's the entire point.
- **It's expected to change** as monitoring gets more correct, and dev tooling shouldn't be pinned to it.

A development session takes precedence over the probe bypass and clears `is_probe`, so a request can never end up half of each.

## Safety

The secret allows impersonating any user with no authentication, so the app **refuses to start** when it's set in a deployed environment rather than trusting it was only ever configured locally. There's a test for that, and one asserting the guard keys on the variable rather than merely on being deployed.

## Also here

`GE_FEED_TRANSPARENCY_FEED` makes the transparency UI's target feed configurable (default unchanged). It hardcoded `your-feed`, which depends on the Perspective API and the trained ranking models and so can't run on a developer machine — leaving the UI empty even once snapshots exist. The dev environment points it at `popularity`.

## Verification

799 tests pass, pyright clean. Verified end to end in the dev environment, not just by unit test: `devctl feed popularity` takes `feed_snapshots` from 0 to 1, `/api/feeds` then lists it, and the frontend renders the real seeded posts that were served (`meidastouch.com`, `weratedogs.com` — from the prod-sampled fixture).

### DIFF
--- a/src/app/main.py
+++ b/src/app/main.py
@@ -61,6 +61,21 @@ from starlette.routing import BaseRoute, Match
 from starlette.types import Scope
 
 
+def _reject_dev_session_secret_in_deployment() -> None:
+    """Refuse to run with the dev-session escape hatch enabled outside local dev.
+
+    GE_DEV_SESSION_SECRET lets a caller assume any user's identity without an
+    AT Protocol JWT (see routers.xrpc.dev_session_did). That is only ever
+    acceptable locally, so refuse to start rather than serve deployed traffic
+    with it enabled — a misconfiguration here is silent and total.
+    """
+    if os.environ.get("GE_DEV_SESSION_SECRET") and _is_deployed_environment():
+        raise RuntimeError(
+            "GE_DEV_SESSION_SECRET must not be set in a deployed environment: "
+            "it allows unauthenticated impersonation of any user"
+        )
+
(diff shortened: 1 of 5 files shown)


raindrift:
Merging this since I need it to move devenv work forward

PR #306, opened 2026-07-23
Terminal feed viewer for the dev environment (#285)

### LINKED ISSUE #285: Dev environment: Simple terminal Bluesky client for viewing dev feeds
Part of the dev environment work (#268). To make generated feeds actually viewable, we want a simple local Bluesky client: request a feed from the local api and display the **hydrated** posts in the terminal. No interactivity or feature set beyond that — it just needs to let a dev see the feed they're working on without publishing anything to Bluesky.

## Scope

- Calls `getFeedSkeleton` on the local api as a chosen dev persona (the seeded DIDs from #268 milestone 1) for a named feed.
- Hydrates each post from the environment's local Elasticsearch (author handle/DID, text, createdAt, like count) — no dependency on the public AppView, so it works entirely against seeded data.
- Prints a readable feed to the terminal: rank, author, timestamp, post text, and the ranker score when the api exposes it.
- Cursor/pagination support so the dev can page through the feed the way a real client would.
- Start from the existing `api/scripts/feed_debug.py` rather than from scratch, and expose it as `devctl feed <feed-name> --user <persona>`.

This supersedes the `devctl feed` bullet in #269 — that milestone can consider the viewer out of scope and depend on this issue instead.

## Non-goals

- Posting, liking, auth against real Bluesky, or any write path.
- A GUI — terminal output is fine. (A web viewer can ride along with the frontend profile later if we want one.)

raindrift:
Closes #285.

> **Stacked on [#303](https://github.com/greenearth-social/api/pull/303)** (`raindrift/dev-session`). Review/merge that first; the viewer authenticates via the dev-session header #303 adds. Once #303 lands this retargets to `main`.

Adds `scripts/feed_view.py`, a read-only terminal client for viewing a locally-generated feed. It requests `getFeedSkeleton` from a running api as a chosen dev persona, hydrates each post from the environment's **own Elasticsearch** — no AppView call, so it works entirely against seeded fixture data — and prints author, timestamp, like count, text, and per-post pipeline detail (which generator retrieved the post, its rank, and the ranker score).

The pipeline detail comes from the feed snapshot the api records on every load, located via the request id carried in each item's `feedContext` token. Flags: `--json` falls back to the raw skeleton (the previous `devctl feed` output), `--no-pipeline` shows the feed as a plain client would, `--pages`/`--cursor` page through it.

`feed_debug.py` and `feed_view.py` render the same concepts (timestamps, scores, media badges), so those formatters move to a shared `scripts/feed_format.py` and both call it rather than drifting apart.

Driven from the dev environment as `devctl feed <feed> [--user <did>]` (internal-tools PR).

Tests: `feed_view_test.py`, `feed_format_test.py`, plus the existing `feed_debug_test.py` still passing over the refactor — 837 total green.

### DIFF
--- a/scripts/feed_debug.py
+++ b/scripts/feed_debug.py
@@ -28,7 +28,7 @@ import argparse
 import asyncio
 import os
 import sys
-from datetime import datetime, timezone
+from datetime import datetime
 from typing import NoReturn
 
 from rich import box
@@ -37,7 +37,12 @@ from rich.panel import Panel
 from rich.table import Table
 from rich.text import Text
 
+sys.path.insert(0, os.path.dirname(__file__))
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "src"))
+
+from feed_format import fmt_score as _fmt_score
+from feed_format import media_badges
+from feed_format import relative_time as _relative_time
 
 from app.documents import FeedDebugDocument
 from app.lib.firestore import (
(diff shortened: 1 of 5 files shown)


raindrift:
This is just a debug utility, and doesn't touch the serving path. Please don't spend time reviewing it.

PR #311, opened 2026-07-24
Fix slow ES serving-path queries behind feed timeouts

### LINKED ISSUE #310: Investigate serving path timeouts and degraded feeds
Now that we have reduced our timeouts to be less than the bluesky client timeout, we are returning a lot of degraded feeds. Some parts of our serving path are slow. Probably ES, but let's verify.

There are a few parts to this story:

1. Look at our metrics to determine which components (CGs, ranker, perspective, etc) are timing out often, or are otherwise unexpectedly slow.
2. If it's obviously ES, check the ES slow query logs on production and see if there are ways to optimize those queries to get us below the timeout. Our target, at this point, is just to return results in less than the timeout (p99). It is ok to run this testing against prod es.
3. Profile the api/inference services in the dev env to see if there is significant savings to be had there.
4. Check that we have appropriate indices for our firebase queries

We did some es index restructuring a while back in order to make our queries faster, mainly splitting up the posts index into separate posts/replies indices, and further dividing it by time. It's possible that we haven't done all the necessary query rewrites to fully take advantage of that restructuring.

For fixes that require a lot of work, or that require schema changes, make new tasks. For fixes that are relatively quick (like query rewrites), go ahead and do them as part of this task. But make sure to test them well, using prod es and a few different users, before deploying.

raindrift:
Closes #310 (the quick-fix portion — follow-up tasks filed separately).

## Motivation

Since feed timeouts were tightened below the Bluesky client timeout, **~8-10% of your-feed loads serve degraded** — ~666 of 7.9k over the last 72h (8.4%), and 8-13% on individual days once the `feed.render.degraded_count` metric began emitting on 2026-07-24. (An earlier revision of this description said ~2%; that was wrong — it divided a partial first-day count of the just-deployed metric by a fuller-window success denominator. See the thread below.) Prod metrics show the rankers, inference, and Perspective all comfortably inside their budgets (p99 < 1s); the degradation is almost entirely ES-side candidate generation — dominated by `two_tower` generator timeouts (686 in 72h), which map ~1:1 onto degraded renders. Component timeouts (4s generator, 2.5s ranker, 1.5s hydration) cap each render, so degradation surfaces as reduced-quality 200s rather than 504s (your-feed render p95≈4.5s, p99≈5s; ~0 hard 504 timeouts).

## What was slow (measured against prod ES)

- **two_tower kNN**: the `like_count >= 20` filter matches ~928k of 20.2M docs in `posts_recent`. A filter that selective makes Lucene abandon the HNSW graph and brute-force scan every matching vector (confirmed via `profile: true` — ~460k vector ops/query). Cold, that's a ~30s query; prod abandons it at the 4s generator timeout while ES keeps grinding.
- **likes lookups**: `likes` docs are indexed with `routing=author_did`, but reads never passed routing — every recent-likes lookup fanned out to all 56 primary shards of the 1B-doc alias.
- **popularity / random_posts / followed_users**: exclusions were handled by overfetching `num_candidates + len(exclude_uris)` (≈2000 docs as seen-lists grow) and filtering client-side.

## Changes

- two_tower kNN: add a 96h `created_at` bound to the kNN filter. The scan set drops to ~320k docs (~95k/day pass the like filter) and is identical for every user, so it stays page-cache hot: 100–350ms warm measured on prod (vs 30s cold before); a fully cold scan re-warms the shared set in one query. Also drop kNN `request_timeout` 60s → 10s so abandoned queries stop burning cluster resources.
- likes fetchers: pass `routing` on the queried DIDs (1 shard per weekly index instead of 7).
- popularity / random_posts / followed_users: push exclusions into the query as `must_not` terms, fetch exactly `num_candidates` (followed_users at prod sizes: **407ms → 29ms**). random_posts keeps sampling the full 2-week alias — downstream features depend on whole-window draws.

## Testing

- Unit suite: 863 passed.
- End-to-end against prod ES (read-only key) + stage inference + live Perspective, for three users including a heavy real user: your-feed first page, cursor paging, past-end regeneration with exclusions, and random feed — all fully populated, kNN at 200–800ms, zero ES timeouts or errors.
- kNN window sizing measured at 48/72/96/120h on prod: counts 136k/230k/322k/411k; warm scans 90–350ms at every size, so 96h buys a 4× candidate pool for ~2× the cold-scan work vs 48h.

### DIFF
--- a/src/app/lib/elasticsearch.py
+++ b/src/app/lib/elasticsearch.py
@@ -87,6 +87,10 @@ async def fetch_recent_liked_post_uris(
                 size=limit,
                 sort=[{"created_at": "desc"}],
                 _source=["subject_uri"],
+                # Likes are indexed with routing=author_did (see ingex), so
+                # routing on the same DIDs searches one shard per weekly index
+                # instead of fanning out to every shard of every index.
+                routing=",".join(user_dids),
             )
 
             data = unwrap_es_response(resp)
@@ -147,6 +151,8 @@ async def fetch_recent_liked_post_uris_and_times(
                 size=limit,
                 sort=[{"created_at": "desc"}],
                 _source=["subject_uri", "created_at"],
+                # Same routing rationale as fetch_recent_liked_post_uris.
+                routing=",".join(user_dids),
             )
 
             data = unwrap_es_response(resp)
--- a/src/app/lib/candidates/es_candidates.py
+++ b/src/app/lib/candidates/es_candidates.py
@@ -21,16 +21,27 @@ async def knn_search_posts(
     exclude_uris: list[str] | None = None,
     ge_post_embedding_model_uuid: str | None = None,
     min_like_count: int | None = None,
+    max_age: str | None = None,
 ) -> list[CandidatePost]:
     """Run a kNN search against the ``posts_recent`` index and return candidate posts.
 
     Filters are passed inside the kNN clause so ES applies them during HNSW
     traversal. The ``posts_recent`` index contains only top-level posts (no
     replies), so no reply-exclusion filter is needed.
+
+    ``max_age`` (an ES date-math duration like ``"48h"``) bounds candidates to
+    recently created posts. When a selective filter such as ``min_like_count``
+    is present, Lucene abandons the HNSW graph and brute-force scans every
+    vector matching the filter — bounding the scan set by recency is what keeps
+    that affordable (and keeps the scanned pages hot across requests, since the
+    set is the same for every user).
     """
     filters: list[dict] = []
     if video_only:
         filters.append({"term": {"contains_video": True}})
+
+    if max_age:
+        filters.append({"range": {"created_at": {"gte": f"now-{max_age}"}}})
 
     if ge_post_embedding_model_uuid:
         filters.append({"term": {"ge_post_embedding_model_uuid": ge_post_embedding_model_uuid}})
@@ -67,7 +78,11 @@ async def knn_search_posts(
             knn=knn_clause,
             size=num_candidates,
             _source=CANDIDATE_SOURCE_FIELDS,
-            request_timeout=60,
+            # Callers abandon this search after a few seconds (the candidate
+            # generator timeout), but ES keeps executing abandoned queries to
+            # completion — a long client timeout here lets a pathological query
+            # burn cluster resources long after anyone is waiting for it.
+            request_timeout=10,
         )
 
     return candidate_posts_from_es_response(resp, generator_name=generator_name)
--- a/src/app/lib/candidates/followed_users.py
+++ b/src/app/lib/candidates/followed_users.py
@@ -106,8 +106,10 @@ async def _followed_users_search_details(
             ],
         }
     }
-
-    fetch_size = num_candidates + len(exclude_uris or [])
+    # Exclusions in the query (not client-side after an overfetch) keep the
+    # fetch size at num_candidates even when a user's seen list is large.
+    if exclude_uris:
+        query["bool"]["must_not"] = [{"terms": {"at_uri": exclude_uris}}]
 
     async with timed(
         logger,
@@ -118,13 +120,14 @@ async def _followed_users_search_details(
         resp = await es.search(
             index="posts_recent",
             query=query,
-            size=fetch_size,
+            size=num_candidates,
             sort=[{"created_at": "desc"}],
             _source=CANDIDATE_SOURCE_FIELDS,
         )
 
     candidates = candidate_posts_from_es_response(resp, generator_name=generator_name)
     if exclude_uris:
+        # ES already excluded these; kept as a cheap safety net.
         exclude_set = set(exclude_uris)
         candidates = [c for c in candidates if c.at_uri not in exclude_set]
     candidates = candidates[:num_candidates]
--- a/src/app/lib/candidates/followed_users_test.py
+++ b/src/app/lib/candidates/followed_users_test.py
@@ -146,7 +146,7 @@ class TestFollowedUsersSearch:
         assert {"term": {"contains_video": True}} not in filters
 
     @pytest.mark.asyncio
-    async def test_exclude_uris_overfetches_and_filters_in_python(self, monkeypatch):
+    async def test_exclude_uris_pushed_into_query(self, monkeypatch):
         stub_followed_dids(monkeypatch, ["did:plc:follow1"])
         es = FakeEs(responses={
             "posts_recent": {
@@ -155,10 +155,6 @@ class TestFollowedUsersSearch:
                         {
                             "_score": 1.0,
                             "_source": {"at_uri": "at://post/1", "content": "x", "embeddings": {}},
-                        },
-                        {
-                            "_score": 0.9,
-                            "_source": {"at_uri": "at://post/excluded", "content": "x", "embeddings": {}},
                         },
                         {
                             "_score": 0.8,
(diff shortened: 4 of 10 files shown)


davidfreifeld:
"Since feed timeouts were tightened below the Bluesky client timeout, ~2% of your-feed loads served degraded (322 of 15.3k over 72h)." -> Out of curiosity, is it really only 2%? The dashboard makes it look like it's more than that:

Screenshot 2026-07-24 at 8 23 31 PM

I'm new to these metrics so this is mostly for my understanding.

src/app/lib/elasticsearch.py
```diff
@@ -87,6 +87,10 @@ async def _fetch() -> list[str]:
                 size=limit,
                 sort=[{"created_at": "desc"}],
                 _source=["subject_uri"],
+                # Likes are indexed with routing=author_did (see ingex), so
+                # routing on the same DIDs searches one shard per weekly index
+                # instead of fanning out to every shard of every index.
+                routing=",".join(user_dids),
```
davidfreifeld, line 93: Again this is mostly for my own understanding. When user_dids is exactly one user_did, then it will only need to look at one shard per weekly index. If it is a list, then it might be multiple shards per index. Though this function as it stands is only ever called with one user_did currently. Is that right?

davidfreifeld:
I just left 2 questions that are mostly for my own understanding. But it looks good to me - excited to see these perf improvements hit!

src/app/lib/elasticsearch.py
mjmor: Yes that's right. In https://github.com/greenearth-social/ingex/pull/412 I found that as little as ~20 DIDs causes basically all shards to be touched. Hence why we need to reduce shard count.

mjmor (approved):

davidfreifeld (approved):

raindrift:
> "Since feed timeouts were tightened below the Bluesky client timeout, ~2% of your-feed loads served degraded (322 of 15.3k over 72h)." -> Out of curiosity, is it really only 2%? The dashboard makes it look like it's more than that:
> 

I think you're right and it's way more. If it was only 2% we wouldn't care. I had the model look at the logs, and I think it was counting log lines incorrectly. I'm not sure why it got such a low number, but I should probably push back on it a little.

raindrift:
Merged latest `main` (resolved conflicts with #309, which landed a per-request `max_age_hours` freshness window through all generators). How the two overlap was reconciled:

- **two_tower recency**: #309 already threads `max_age_hours` into the kNN filter — the same mechanism this PR introduced. But the default is 168h (7d), which is exactly the ~928k-doc / ~30s brute-force-scan case #310 is about. So instead of a separate hardcoded window, two_tower now **clamps** the requested window: `min(max_age_hours, TWO_TOWER_MAX_AGE_CAP_HOURS)` with the cap at 96h. Freshness presets ≤72h pass through unchanged; only the 7-day default is bounded (to keep the brute-force vector scan hot and affordable). `knn_search_posts`'s `max_age_hours` was widened `MaxAgeHours → int` since the cap (96) is deliberately off the user-facing freshness ladder.
- **exclusion push-down** (popularity / followed_users / random_posts): re-applied on top of #309's versions — `must_not` in the query, fetch exactly `num_candidates`.
- **random_posts**: kept #309's freshness filter (per discussion, random_posts honors `max_age_hours` like the other generators); this PR only adds the exclusion push-down there.
- **likes routing** and **kNN `request_timeout` 60s→10s**: unchanged from before, no conflict (main didn't touch `elasticsearch.py`).

Re-verified against prod ES post-merge: your-feed + random served for the probe and a heavy real user, two_tower kNN at ~160ms with the 96h cap, no timeouts. Full suite: 888 passed.

raindrift:
## Prod-ES validation (2026-07-27, post-merge)

Replayed the exact query shapes the merged code produces, cold (`request_cache=false`), against prod ES with real data and real users, comparing old vs new. **Every serving-path query now clears the 4s generator timeout and 9s feed budget comfortably, even cold.**

### A. two_tower kNN — the ~3.8% timeout driver (headline win)

`like_count>=20` + model_uuid + a realistic 1500-URI exclusion list. 3 cold runs each:

| variant | took ms (min/med/max) | vs 4s generator timeout |
|---|---|---|
| pre-#310, no window (original bug) | 234 / 890 / **11685** | blows it |
| main today (#309, 168h window) | 334 / 556 / 1313 | under, but thin |
| **merged, 96h cap (#311)** | **50 / 55 / 575** | ~7x margin |

All variants returned a full 30 candidates — the tighter window is not starving the slate. #309's 168h filter already recovered most of the ground (11.7s → 1.3s max); the 96h cap here takes the worst case to 575ms.

### B. recent-likes routing (5 real high-activity users, up to 279k likes each)

Clear win on the heavy users that drive tail latency; lighter users land in the cold-cache noise band but all under 700ms:

```
yf4c7cxemwekc752   no-routing 669ms -> routing 464ms  (hits=50)
htpkpg4fvpif667d   no-routing 420ms -> routing 122ms  (hits=50)
nbm6mxto6mknrluh   no-routing 173ms -> routing 225ms  (hits=50)
ifxftkewd54f2oqk   no-routing 112ms -> routing 342ms  (hits=50)
sd5n4rk3aeosd6j4   no-routing  82ms -> routing 190ms  (hits=50)
```

Routing's larger benefit — cutting shard fan-out from 7→1 per weekly index — reduces cluster-wide load in a way isolated single-query timing understates.

### C. exclusion push-down (big 1500-URI seen-list)

New `must_not` + `size=30` vs old overfetch `size=1530` + client-side filter. 3 cold runs:

| variant | took ms (min/med/max) |
|---|---|
| OLD overfetch size=1530 | 733 / 753 / 1705 |
| NEW must_not size=30 | 573 / 657 / **705** |

Faster, and a much tighter tail (705ms vs 1705ms) by cutting the fetch phase.

### Caveat

Much of the win is bounded-scan behavior that stays cheap because the scanned set is shared across users and stays page-cache-hot. If the cluster RAM pressure in #312 worsens, cold re-warms would get slower — but even the cold numbers above clear the budget.

raindrift:
(I asked Claude to check the metrics again, and here's what it said)

You're right, and thanks for pushing on it — the 2% was wrong. The real degraded rate is **~8-10%**. I've corrected the PR description.

**What happened:** `feed.render.degraded_count` only started emitting in prod on **2026-07-24** — the exact day I pulled the original numbers. Daily your-feed degraded_count:

```
2026-07-21     0        (metric not yet emitting)
2026-07-22     0
2026-07-23     0
2026-07-24   313   10.1%   <- my original pull landed here, mid-first-day
2026-07-25   175    7.9%
2026-07-26   290   13.1%
2026-07-27   140    4.4%  (partial day)
```

My "322 / 15.3k = 2%" divided a **partial first-day count** of the freshly-deployed metric by a fuller-window success denominator. Over the last full 72h it's **666 / 7906 = 8.4%**, and it's 8-13% on complete days.

**Cross-check that confirms ~8-10% is real:** the degradation is almost entirely candidate-generator timeouts, dominated by `two_tower` (686 timeouts / 72h). Those map ~1:1 onto degraded renders (666), so the two independent signals agree. I had the 553 two_tower-timeout number in the original writeup but wrongly treated it as a footnote rather than recognizing it *is* the degradation rate.

**Two more corrections while I was in there:**
- These surface as reduced-quality **200s, not 504s** — the per-stage timeouts (4s generator / 2.5s ranker / 1.5s hydration) cap each render at ~5s (your-feed p95≈4.5s, p99≈5s), so the 9s pipeline timeout almost never fires (~0 504s). So "degraded" is the right primary signal, not timeouts.
- This makes the fix *more* important, not less: two_tower timeouts are the degradation, and the 96h cap takes two_tower kNN worst-case from 11.7s → 575ms (see the validation comment above), which should pull the 8-10% down substantially. The fix direction is unchanged and validated — only the severity framing was off.

Worth adding a dashboard panel for degraded-rate = degraded_count / success_count per feed so this is visible going forward; I'll file that.

Issue #189, opened 2026-06-24
Simulated user load testing


raindrift:
We should pick a random set of real users, and then simulate load from them.

We need to be able to do this arbitrarily, by specifying an amount of load and an amount of time. We eventually want to automate this kind of test periodically until we have enough users that our real traffic does it for us, but the automation is a separate story.

We should tag these in firebase so we can filter them out later, and also in any metrics we log. We will need to bypass auth (probably using an api token) to do this, so we'll know in the backend whether it's a test. So we can use that info to tag records.

These can be generic bsky users, or users of our feed. We probably want a mix of them, so that we will be hitting our cold-start paths. Cold-start can be a little slow or degraded for a new user, and that's fine, but since we are planning to scale up relatively rapidly, we want to make sure cold users don't cause our infra to fall over.

Important things:
- tests should be tagged in all metrics, firebase, etc
- we have to be careful not to break our real user numbers in (eg) posthog. Probably we should skip logging to posthog for these tests entirely.
- If a user was created entirely from a load test (like we have never seen real traffic from them), it should be possible to remove their data from firebase so that we can re-run a test with the same user and get the cold path again. If they visit for real though, they become our user and we should not delete their data (fyi, I am mostly unconcerned about race conditions here because they'd be quite rare)

Let's do this for 100 users to start? They need to be a diverse set of profiles, with lots of different numbers of followers, levels of activity, etc.

Once we have the script, run one test at 5x our current scale and see what breaks. If nothing breaks, keep increasing it until something does. We should probably perform this testing during a low-traffic period, since it will at least briefly disrupt prod.

Here are the steps:
1. Fix all the metrics logging and db writes so they tag the data properly
2. Write the script that performs load testing with some reasonable set of configurations, and at a settable rate
2. Run analysis to find users for load testing. 
4. Run the test
5. Analyze to see if anything in the serving path broke, if latency got a lot higher, and if so where

We should have scripts for finding users to test with, for running the test, and for analyzing the results. The analysis script should look at relevant metrics (latency, failures, etc). Probably we want to have the load generator and analysis be separate steps, even if one generally runs right after the other.

If it makes sense for this to run in devenv (for convenience with dependency management) that is ok. But let's not make a new devctl command for it. Running it with exec is fine. Running it locally (without the devenv) is also fine. Whichever seems best.

raindrift:
[...]

PR #315, opened 2026-07-25
Simulated user load testing with isolated test traffic (#189)

raindrift:
Adds the ability to simulate feed load from a diverse set of real Bluesky users, with backend changes that keep all test-generated data, metrics, and analytics fully isolated from real users — so we can find out what breaks under scale before real growth does.

> ⚠️ **Security:** `GE_LOAD_TEST_SECRET` lets a caller impersonate any `did:plc:` user against deployed environments (that's the point — the load has to hit prod). It's the same power the dev-session bypass is *banned* from prod for. Mitigations: secret lives only in Secret Manager, every record it produces is tagged as test data or is pipeline-equivalent to a real request, each load-test request logs a warning, and the tagged data is deletable. Rotate the secret after each campaign.

Deploy the backend and run `gcp_setup.sh` once per env (creates `load-test-secret-<env>`) before the first run.

Adds simulated user load testing: drive \`getFeedSkeleton\` (with cursor paging and \`sendInteractions\`) for a chosen set of real user DIDs, while keeping all resulting data, metrics, and analytics isolated from real users. Goal: learn whether the serving path — especially cold-start — falls over under 5x+ current load before real growth does it for us.

**Security:** \`GE_LOAD_TEST_SECRET\` permits impersonating any \`did:plc:\` user in deployed environments (required — the load must hit prod). Same power the dev-session bypass is banned from prod for. Mitigations: Secret Manager only, all resulting data tagged as test data or pipeline-equivalent to a real request, a per-request warning log, and deletable tagged data. Rotate the secret after each campaign.

### Backend
- **New auth mode** \`load_test_did\` (\`GE_LOAD_TEST_SECRET\` + \`X-Load-Test-Secret\` / \`X-Load-Test-DID\`). Allowed in deployed envs, unlike the dev session. Precedence: dev session > load test > probe > JWT. Treated as a real session downstream (writes snapshots, feed cache, discarded posts) so the cold-start path is exercised for real.
- **Tagging:** \`UserDocument.created_by_load_test\`, \`InteractionDocument.load_test\`, \`FeedCacheDocument.load_test\`, and a feedContext \`lt\` claim (so the public sendInteractions endpoint recognises test interactions without client cooperation). \`upsert_user\` tags newly created users, never mutates an existing user from test traffic, and clears the flag on the first real request (the user "becomes ours" and is no longer deletable).
- **Metrics:** request-scoped \`traffic\` label (\`real\` | \`probe\` | \`load_test\`) auto-attached to every metric via the existing request_context ContextVar (propagates into background tasks).
- **Analytics:** PostHog session/interaction tracking skipped for load-test traffic; load-test "seen" events never enter real users' seen-post exclusions.

### Scripts (\`api/scripts/\`)
- \`load_test_users.py\` — cohort selection (default 60% existing feed users / 30% active likers from ES / 10% cold), writes a users JSON.
- \`load_test.py\` — async Poisson-paced session generator with paging + interactions and a rate guardrail; writes per-request JSONL.
- \`load_test_analyze.py\` — client-side percentiles plus Cloud Monitoring (split by \`traffic\`) and Cloud Run log counts. **Reads no Firestore**, so cleanup can run before analysis.
- \`load_test_cleanup.py\` — deletes flagged users (recursive), interactions, and cache entries; dry-run by default, re-checks the flag before each delete to avoid the become-real race.

### Deploy
\`gcp_setup.sh\` / \`deploy.sh\` provision and wire \`load-test-secret-<env>\`.

### Testing
Full api suite green (913 passed). New coverage for the auth mode, snapshot-write contrast with the probe, upsert_user semantics, feedContext \`lt\` backward-compat, the metric label, interaction provenance, and the script helpers/cleanup logic.

Follow-up (not in this PR): run \`gcp_setup.sh\` per env, then the operator runbook in the issue — select users, run at ~5x during a low-traffic window, analyze, escalate until something degrades. After the label ships, add \`metric.label.traffic != "load_test"\` filters to any console dashboards/alerts that should ignore test traffic.

### DIFF
--- a/scripts/load_test/analyze.py
+++ b/scripts/load_test/analyze.py
@@ -99,7 +99,54 @@ def _client_summary(records: list[dict]) -> None:
     )
 
 
-def _server_metrics(project: str, start: datetime, end: datetime) -> None:
+# OTel resource attribute service.namespace (set to the environment in
+# metrics.py) surfaces as this monitored-resource label in Cloud Monitoring.
+# It separates stage from prod, which share the project and the metric type.
+ENV_RESOURCE_LABEL = "namespace"
+
+# The percentiles to pull, mapped to Cloud Monitoring per-series aligners. The
+# distribution metric is reduced server-side, so we never reconstruct it
+# client-side (no cumulative double-counting, no mean-of-means).
+_PERCENTILE_ALIGNERS = {
+    50: "ALIGN_PERCENTILE_50",
+    95: "ALIGN_PERCENTILE_95",
+    99: "ALIGN_PERCENTILE_99",
+}
+
+
+def build_percentile_request(
+    monitoring_v3,
+    project: str,
+    metric_type: str,
+    env: str,
+    interval,
+    percentile: int,
+    alignment_seconds: int,
+):
+    """Build a ListTimeSeries request for one percentile, grouped by traffic class.
+
+    Server-side aggregation aligns each series to the requested percentile over
+    the whole window, then reduces across series (e.g. per-instance) within each
+    ``traffic`` label. Filtered to one environment so stage and prod — which
+    write the same metric type to the same project — are never mixed.
+    """
+    aligner = getattr(monitoring_v3.Aggregation.Aligner, _PERCENTILE_ALIGNERS[percentile])
+    aggregation = monitoring_v3.Aggregation(
+        alignment_period={"seconds": alignment_seconds},
+        per_series_aligner=aligner,
+        cross_series_reducer=monitoring_v3.Aggregation.Reducer.REDUCE_MEAN,
+        group_by_fields=["metric.label.traffic"],
+    )
+    return monitoring_v3.ListTimeSeriesRequest(
+        name=f"projects/{project}",
+        filter=f'metric.type = "{metric_type}" AND resource.labels.{ENV_RESOURCE_LABEL} = "{env}"',
+        interval=interval,
+        aggregation=aggregation,
+        view=monitoring_v3.ListTimeSeriesRequest.TimeSeriesView.FULL,
+    )
+
+
+def _server_metrics(project: str, env: str, start: datetime, end: datetime) -> None:
     try:
         from google.cloud import monitoring_v3
     except ImportError:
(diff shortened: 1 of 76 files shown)


raindrift:
I believe this covers all the code changes needed to be able to do load tests while keeping track of the source of the traffic, and cleaning up firebase afterward. Obviously I haven't run the tests yet, because we need to merge and deploy all of this first.

src/app/routers/xrpc.py
davidfreifeld: I believe this will add low score posts to exclusions for both real and load test users? So we would probably want to skip it for load test users.

scripts/load_test/lib.py
```diff
@@ -0,0 +1,181 @@
...
+    Geometric distribution with the given mean, so most sessions are short and a
+    few page deep — the shape real feed sessions tend to have.
+    """
+    if mean_pages < 1:
+        raise ValueError("mean_pages must be >= 1")
+    # Geometric with support {1, 2, ...} has mean 1/p, so p = 1/mean_pages.
+    p = 1.0 / mean_pages
+    # inverse-CDF sample; clamp the random draw away from 0 and 1.
+    u = min(max(rng.random(), 1e-9), 1 - 1e-9)
+    depth = 1 + int(math.log(1 - u) / math.log(1 - p))
```
davidfreifeld, line 152: If mean_pages is 1 here, we'll try to take log(0)

scripts/load_test/analyze.py
```diff
@@ -0,0 +1,234 @@
...
+    metric = f"{METRIC_PREFIX}/feed.render.duration_ms"
+    request = monitoring_v3.ListTimeSeriesRequest(
+        name=f"projects/{project}",
+        filter=f'metric.type = "{metric}"',
+        interval=interval,
+        view=monitoring_v3.ListTimeSeriesRequest.TimeSeriesView.FULL,
+    )
+    by_traffic: dict[str, list[float]] = defaultdict(list)
+    try:
+        for series in client.list_time_series(request=request):
```
davidfreifeld: A few things here (I'm not 100% on these so check me): 

- Each `point` is a cumulative metric. So when we extend `by_traffic` we're counting earlier points multiple times
- Do we need to take the mean here? I think we can just use the histogram values already available, and get a more representative distribution.
- Might we be mixing stage and prod metrics together instead of keeping them separate?

src/app/lib/firestore.py
```diff
@@ -118,6 +120,21 @@ async def upsert_user(db: AsyncClient, user_did: str, username: str | None) -> U
...
+      ``last_seen_at`` bump. A load test samples real users, and bumping their
+      activity on every run would corrupt the very activity data the user
+      selection reads. The read already happened; skipping the write is free.
+
+    Real traffic (``is_load_test=False``) on a document previously created by a
+    load test clears the tag and resets ``created_at`` to now: a real person has
+    now shown up, so the user "becomes ours", their data must be preserved, and
+    their first-seen time should reflect the first *real* visit.
     """
     ref = db.collection(USERS_COLLECTION).document(user_doc_id(user_did))
```
davidfreifeld, line 147: Outside of my real area of expertise but (thanks Codex):

The “synthetic user becomes real” invariant is racy. upsert_user performs a read followed by an unconditional set (firestore.py:139), (firestore.py:181) If a real and load-test first visit both observe a missing document, the load-test write can land last and mark the real user deletable. Cleanup’s re-read followed by recursive_delete is another TOCTOU window (cleanup.py:93): a real request can clear the flag after the check but before deletion. These transitions need transactional/precondition-based protection.

src/app/routers/xrpc.py
davidfreifeld, line 117: (From Codex): 

Initial-request reuse can cross between real and load-test traffic. The five-second reuse key contains only DID, feed, and limit (xrpc.py:109); is_load_test is omitted when it is claimed (xrpc.py:1478). For an existing user, a real and simulated request can therefore share a response carrying the leader’s lt claim. If the real request leads, simulated interactions are treated as real—including seen-post writes and PostHog tracking. If the load test leads, genuine interactions are discarded from those systems. Traffic mode must be part of the reuse key.

davidfreifeld, line 1645: (From Codex):

Load-test snapshots pollute the user-facing transparency feed. Load tests pass the not is_probe snapshot-writing guard (xrpc.py:1532), but FeedSnapshotDocument has no test provenance field (documents.py:406). /api/feeds then returns those snapshots without filtering (feed_transparency.py:172). For existing users, cleanup cannot identify or remove them, so synthetic loads can fill their recent-feed UI for the 15-minute window.

scripts/load_test/cleanup.py
```diff
@@ -0,0 +1,161 @@
...
+    user_doc_id,
+)
+from load_test.lib import GCP_PROJECT
+
+console = Console()
+
+_FIRESTORE_DATABASES = {"stage": "greenearth-stage", "prod": "greenearth-prod"}
+
+
+def _configure_environment(env: str) -> None:
```
davidfreifeld, line 65: (From Codex):

Cleanup’s safety flags do not enforce their advertised scope. Selecting dev simply preserves inherited environment variables without requiring an emulator (cleanup.py:51); with --execute, absent emulator configuration could target credentials/default Firestore instead. Also, --users restricts only user-document deletion—interactions and caches are still deleted globally (cleanup.py:127). The CLI should fail closed for dev and either honor the DID restriction everywhere or describe its actual scope.

src/app/lib/firestore.py
raindrift: Thanks for calling this out. I thought about this race, and explicitly decided to avoid protecting against it. The chance that a specific individual user shows up to join at precisely the moment when we are creating their fake account during a load test seems so vanishingly unlikely, that the small inconsistency is ok. Like, worst-case, their data gets deleted and then recreated properly on their next feed load. Since the account was new anyway, it's nbd.

raindrift:
Addressed the review comments in d2a530a. Summary of how each was handled:

**Data isolation** — the through-line was that three sibling write paths leaked load-test effects into real users. Rather than *skip* those writes (which wouldn't exercise the paths the load test exists to measure), each now writes but stays isolated:

- **Seen + discarded posts** (@davidfreifeld's first comment): kept the writes, but routed to a separate per-user daily bucket `YYYY-MM-DD-load-test`. The exclusion reader picks up both real and load-test buckets, so exclusion-growth behaviour under test is identical to prod; cleanup can delete exactly the synthetic buckets. This replaces the earlier "skip seen for load test" approach.
- **Feed snapshots** (transparency-pollution comment): write-then-delete for load-test traffic — the transactional write is exercised, but the doc is removed immediately so it never appears in the user-facing `/api/feeds` transparency view. Snapshots are read only there, so no provenance field or reader-side filter was needed.
- **Initial-request dedup** (reuse-key comment): added the traffic class to the 5s reuse key so a real and a load-test request for the same `(user, feed, limit)` never share a response (which would hand the follower the leader's `feedContext` `lt` provenance). This was a pre-existing thundering-herd guard; attribution stays purely auth-based.

**Bugs:**
- `sample_page_depth` crashed at `--mean-pages 1` (`log(0)`) — short-circuits to 1.
- `analyze.py` server metrics reworked to server-side `ALIGN_PERCENTILE` aggregation grouped by the `traffic` label and filtered to one environment's resource label — fixes the cumulative double-counting, the mean-of-means, and stage/prod mixing you flagged.

**Cleanup hardening** (safety-flags comment): fail-closed on `--environment dev` unless a Firestore emulator is configured; `--users` now scopes interactions and activity buckets too, while `feed_cache` (no user linkage) is skipped rather than deleted globally; added deletion of real users' load-test activity buckets.

**Not changed** — the `upsert_user` "synthetic user becomes real" race: per my earlier reply, the window is vanishingly unlikely and self-heals on the user's next real feed load, so I've left it.

Also cleaned up a duplicate `posthog_client` import left by the `main` merge. Full suite green (987) and CI (pyright) passing.

PR #318, opened 2026-07-27
Enforce ruff in CI and fix all lint/format violations

raindrift:
## Motivation

The Python CI workflow ran pyright and pytest but **never ran ruff**, so lint and formatting drift landed on `main` unchecked. This aligns the repo's CI with `internal-tools`, the current reference Python CI config.

## What changed

- **CI** (`python-ci.yml`): add `ruff check` and `ruff format --check` steps, and pin `actions/setup-python` to v6 to match the reference.
- **Formatting**: `ruff format` applied across the repo — this is the bulk of the diff and is purely mechanical.
- **Lint fixes**: import ordering (I001), unused imports/vars (F401/F841), `zip(..., strict=False)` (B905), `raise ... from err` in `except` (B904), `StrEnum` for `str`+`Enum` (UP042), specific-exception assert (B017), and line-length (E501) via manual wrapping.
- **E402**: moved imports to module top in two test files; scoped an `E402` per-file-ignore for `main.py`, whose imports intentionally follow root-logger configuration.

## Safety

No behavior changes. B905 fixes preserve current semantics (`strict=False`), and `DegradationStage` is never string-formatted (so the `StrEnum` switch is inert). Locally: `ruff check`, `ruff format --check`, `pyright`, and the full 887-test suite all pass.

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -16,7 +16,7 @@ jobs:
       uses: actions/checkout@v7
 
     - name: Set up Python
-      uses: actions/setup-python@v7
+      uses: actions/setup-python@v6
       with:
         python-version: '3.13'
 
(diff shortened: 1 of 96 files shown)


raindrift:
I wanted to see how much of a mess this would be to try. Turns out it is a lot of a mess. So probably I'll throw this on the back burner until we have a moment when there are few open PRs and nobody is around, like a holiday or something.

PR #325, opened 2026-07-29
api reader migration: use fields API for post embeddings

### LINKED ISSUE #312: ES cluster memory pressure: page-cache churn makes identical queries 100x slower under load
Spun out of #310.

While profiling the serving-path slowness we found the prod ES data nodes running at 88–100% RAM (`_cat/nodes` ram.percent; data-1 at 100%). The symptom is that query cost depends almost entirely on page-cache residency, not query shape:

- The two_tower kNN brute-force scan: ~30s cold vs 26ms warm (same query).
- A trivial likes lookup (terms on author_did, size 50): 27ms when replayed off-peak, but p50 **2.5s** in the prod slow-query logs across many users.
- Trivial `terms at_uri` fetches (size ≤ 100) intermittently taking 0.5–1.5s.
- Slow-query volume runs all day (~100/h over the 500ms threshold) with an 18:00–20:00 UTC spike.

Likely contributors worth investigating:

- The `likes` alias is 1.04B docs across 8 weekly indices × 7 shards × 2 replicas = 112 shard copies, sharing the 4 data nodes with the posts/vector data. It almost certainly evicts the (much hotter) posts_recent vector + doc data.
- 548 total shards on 4 data nodes is a lot of per-shard overhead.
- 7 shards/week for likes looks over-provisioned; the API only ever queries them by author_did (now routed after #310's fix, so 6/7 shards per index sit idle for those reads).

Options: more RAM / dedicated node tiers, shrink likes retention or shard count (ILM/shrink), or move likes to their own tier. Needs sizing work — hence a separate task.

mjmor:
Closes #312

Step 1 of the ES memory/hydration plan from the latest comment on #312: `_source` on the `posts` index is ~65% embeddings by size, so hydrating candidates by `at_uri` pulls far more bytes than needed and churns page cache.

# This PR

- Switch `fetch_post_embeddings` and `fetch_post_embeddings_and_metadata` (the only serving-path readers of embeddings from `_source`) to ES's `docvalue_fields` retrieval for the MiniLM L12 vector, with `_source` now excluding `embeddings` entirely.
- Add a shared `embedding_from_fields` helper (`elasticsearch.py`) that unwraps `docvalue_fields`'s outer list around the dense_vector value.
- Extend the same fix to the `/skylight/search` and `/skylight/similar` debug endpoints (`skylight.py`), which had the same problem — their ES calls had no `_source` filter at all, pulling the entire document (all three embedding families) even though only the MiniLM L12 vector is ever used.
- Confirmed `ge_post_embedding` and `all_MiniLM_L6_v2` don't need the same treatment: `ge_post_embedding` is only ever used as a `knn.field` search target (never read back), and `all_MiniLM_L6_v2` isn't read anywhere in `api` or `inference-service`. Both are safe to drop from `_source` in step 2 with no reader-side changes.

**`fields` vs `docvalue_fields`:** this PR originally used the `fields` retrieval option. Verified live against prod (ES 9.0.0) that `fields` does return the correct vector with `_source: false` — but a report from a reviewer's Codex run benchmarked `fields` as ~19-21% slower end-to-end with visibly higher ES-side p50/p95 than `docvalue_fields`, consistent with `dense_vector` not getting a source-free value fetcher for `fields` until ES 9.2 (i.e. `fields` silently falls back to decompressing `_source` on our cluster today, defeating the point of this migration). `docvalue_fields` always reads from doc values by definition, independent of ES version, so we switched to it. One correctness note along the way: `fields` returns a dense_vector as a flat list, while `docvalue_fields` wraps it in an outer list (`[[...]]`) — the helper and all test fixtures assert the shape actually returned by prod for whichever option is in use.

# Testing

- `pytest src/app` — all 740 tests pass, including new/updated coverage in `elasticsearch_test.py`, `skylight_test.py`, `candidates_test.py`, and `post_similarity_test.py` for the `docvalue_fields` response shape.
- `ruff check` and `pyright` on changed files — clean (no new issues; pre-existing lint debt in `skylight.py`/`skylight_test.py`/`candidates_test.py` untouched).
- Verified live against prod ES 9.0.0 (`posts`, `posts_recent`, old and current weekly indices) that `docvalue_fields` returns the correct 384-float vector with `_source: false`.


davidfreifeld:
Codex claims that with ES 9.0 (the version we're using, right?) - we need to use `docvalue_fields` instead of `fields`. It says that plain `fields` in ES < 9.2 falls back to using _source. And also that `fields` doesn't return a nested list, as assumed by this PR (whereas `docvalue_fields` does). Let me know if you'd like the full Codex output. Otherwise the idea here and the code lgtm!

davidfreifeld (approved):

Issue #343, opened 2026-07-31
Additional loadtesting telemetry


mjmor:
## Problem

The load-test tooling (#189) can generate load and cleanly separate test traffic (the `traffic` metric label), but we can't *assess* the results well. Analyzing the Jul 30 tests surfaced two missing capabilities:

1. **Correlate infrastructure signals with user experience.** Page-cache churn, memory pressure, and IO live in different metric families (Prometheus ES exporter, GKE, GCE) with nothing aligning them to latency/failure/degradation metrics over the same windows.
2. **Disambiguate the bottleneck.** api-service saturation vs ES backend vs downstream (inference-service, Perspective) — the Jul 30 diagnosis (single-instance saturation + autoscaler lag, *not* steady-state capacity) required manual ad-hoc queries and inference across five signals.

Underlying instrumentation gaps: api ES queries are log-only (no metric, and the server-side `took` is discarded); no event-loop lag signal; inference-service has zero server-side metrics; external-call failures carry no status codes.

## Plan

**1. Attribution instrumentation** — small PRs, deployed before the next test ladder. Principle: every dependency call gets a client-side and a server-side duration; the gap is queuing on our side.
   - `es.query.duration_ms` + `es.query.took_ms` with an `op` label (`likes`/`hydrate`/`knn`/`popularity`/`author_scan`), recorded in the `es_client.py` search wrapper. `took` ↑ ⇒ ES is slow; gap ↑ with flat `took` ⇒ api is slow. Also gives #338 its per-workload baseline series.
   - `eventloop.lag_ms` — asyncio scheduling overshoot per instance; the direct "api instance saturated" signal.
   - inference-service server-side `inference.predict.duration_ms` (port of api's metrics module); pairs with client-side `rank.model.duration_ms`.
   - `status_code` labels on Perspective/Bluesky failure counts (429 = throttled vs timeout = our queuing).

[...]

PR #350, opened 2026-08-01
Load-test regression assessment: ES client/server attribution metrics

mjmor:
Part of #343 (PR A of the small-PR series; the final PR in the series closes the issue).

Step 1 of the regression-assessment plan (#343): every ES read gets a client-side and a server-side duration, so "ES is slow" and "the api is slow" are distinguishable from metrics alone. Also instantiates #338 §2's workload table as time series.

# This PR

- Record `es.query.duration_ms` + `es.query.took_ms` (histograms, `op` label) in the `SlowQueryLoggingES` search wrapper; `op` defaults to `unlabeled` so missed call sites surface in Monitoring
- Supply explicit `op` at all 11 ES search call sites: `likes`/`hydrate`/`knn`/`popularity`/`author_scan`/`random`/`search`

# Testing

- `pipenv run pytest -q` — 1030 passed on the branch tip (baseline 991 before this branch)
- New unit tests: wrapper records both metrics with `op`; `op` never forwarded to the ES client; missing `took` tolerated; timeout path records duration only; call sites pass correct `op`
- Audit: `grep -rn "\.search(" src/app` shows no ES call site without `op=`


src/app/lib/candidates/network_likes.py
```diff
@@ -73,6 +73,7 @@ async def fetch_recent_liked_post_uri_page(
 
     resp = await es.search(
         index="likes",
+        op="likes",
```
davidfreifeld, line 76: nit: do we possibly want to distinguish this from "likes" as it's used in elasticsearch.py to look up a user's recent likes as features for the models?

davidfreifeld (approved):

src/app/lib/candidates/network_likes.py
mjmor: I don't think so. This tag explains the pattern of query sent to ES to help identify slow query patterns, the reason the caller wants that data need not be captured.

PR #359, opened 2026-08-03
Annotate PostHog events with surface and schema_version

### LINKED ISSUE #357: Annotate PostHog events with surface and schema_version
The frontend is introducing PostHog instrumentation (greenearth-social/frontend#63) into the same PostHog project this service already writes to. Every frontend event is stamped with `surface: "greenearth_web"` and `schema_version` at a single chokepoint (`PostHogAnalyticsService.capture`), so frontend events are self-identifying. API events currently carry neither, which means the only thing separating the two producers in the shared project is PostHog's automatic `$lib` property (`web` vs `posthog-python`) — an implementation detail of the SDKs rather than a convention we control.

That is workable but fragile. It breaks the moment a third producer appears (a script, a worker, a future service), it can't express schema evolution, and it forces every insight author to know an undocumented SDK quirk.

# Proposal

Stamp the same two annotations onto every event this service emits:

- `surface: "greenearth_api"` — mirrors the frontend's `greenearth_web`, one value per event producer.
- `schema_version: 1` — per-surface, independent of the frontend's version number. Only meaningful in combination with `surface`.

Requirements:

- Applied at a single chokepoint in `src/app/lib/posthog_client.py`, so no call site can forget them.
- Applied *after* caller-supplied properties, so an event property can never silently overwrite the partition key. (The frontend spreads in the opposite order — worth not copying that.)
- `scripts/backfill_posthog.py` emits the same annotations; backfilled events are API-origin and should not be a gap in the partition.

Non-goals: changing event names, changing `distinct_id`, or touching feature-flag evaluation.

mjmor:
Closes #357

The frontend is adding PostHog instrumentation (greenearth-social/frontend#63) that writes to the same PostHog project this service already uses. Frontend events are stamped with `surface: "greenearth_web"` and `schema_version` at a single chokepoint; API events carried neither, so the only thing separating the two producers was PostHog's automatic `$lib` property (`web` vs `posthog-python`) — an SDK implementation detail rather than a convention we control, and one that can't express a third producer or schema evolution.

# This PR

Stamps every event this service emits with `surface: "greenearth_api"` and `schema_version: 1`. Specifically:

- Add `EVENT_SURFACE` / `EVENT_SCHEMA_VERSION` and `annotate_event_properties()` to `src/app/lib/posthog_client.py`
- Apply the annotations inside `track_session` and `track_interaction`, so no call site can forget them
- Apply annotations *after* caller-supplied properties, so an event property named `surface` can't silently reassign the event to another producer — the frontend spreads in the opposite order, which is a footgun worth not copying
- Stamp the same annotations in `scripts/backfill_posthog.py`, so historical API-origin events aren't a hole in the partition
- Document the cross-repo convention in `README.md` (new "Analytics (PostHog)" section), including that event names are *not* namespaced across surfaces so insights must filter on `surface`

`schema_version` is scoped to the surface — the frontend versions its own schema independently, so the number is only meaningful alongside `surface`.

No changes to event names, `distinct_id`, or feature-flag evaluation.

# Testing

- `pipenv run pytest` — 996 passed
- New tests in `src/app/lib/posthog_client_test.py`: annotations present on all four capture paths, caller properties can't override them, and `annotate_event_properties` doesn't mutate the caller's dict
- Updated existing assertions in `posthog_client_test.py` and `scripts/backfill_posthog_test.py` to expect the annotations
- `pipenv run ruff check` / `ruff format --check` on the touched files — clean (the repo has ~408 pre-existing lint errors across 83 files; the one `E402` reported in `posthog_client_test.py` is a pre-existing mid-file import, untouched by this PR)

### DIFF
--- a/README.md
+++ b/README.md
@@ -553,6 +553,34 @@ Disable when done (full capture has storage/perf cost):
 pipenv run scripts/feed_debug.py [username].bsky.social --environment stage --disable
 ```
 
+## Analytics (PostHog)
+
+This service and the frontend write to the **same PostHog project**, so every
+event this service emits is annotated to identify its producer. Annotations are
+applied centrally in `src/app/lib/posthog_client.py` — call sites never set them.
+
+| Property | Value here | Purpose |
+|---|---|---|
+| `surface` | `greenearth_api` | Which producer emitted the event. The frontend stamps `greenearth_web`. |
+| `schema_version` | `1` | Version of *this surface's* event schema; scoped to `surface`, versioned independently of the frontend. |
+
+Conventions:
+
+- **Filter on `surface`** in any insight that should cover one producer rather
+  than both. Event names are not namespaced, so a name collision between the two
+  surfaces is possible and only `surface` separates them.
+- `schema_version` is only meaningful alongside `surface` — the two surfaces'
+  version numbers are unrelated. Bump it when an existing event's properties
+  change shape in a way that would break a saved insight.
+- Annotations are applied *after* caller-supplied properties, so an event
+  property can never overwrite the partition key.
+- `scripts/backfill_posthog.py` stamps the same annotations, so historical
+  API-origin events are not a gap in the partition.
+
+`distinct_id` is the user's `did:plc:…` on both surfaces (the frontend's Firebase
+`uid` is that same DID), so a user is one PostHog person across both. Feature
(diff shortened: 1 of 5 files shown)


Gautham-Raju (approved):
LGTM, thanks!

Issue #344, opened 2026-07-31
API service cannot handle lots of concurrent requests


raindrift:
We ran some load tests today (7/30). We expected to have elasticsearch issues, but actually the API service seemed to be the bottleneck. We were able to handle increasing volumes of traffic by scaling the number of Cloud Run instances, but it was way more than it should have been.

We ran the following load tests, along with the qpm and number of api instances we had serving them:

2026-07-31T02:27:05.175730+00:00 : 150qpm, 1 instance (2 minutes)
2026-07-31T03:03:27.125254+00:00 : 30qpm, 1 instance
2026-07-31T03:16:56.775914+00:00 : 90qpm, 4 instances
2026-07-31T03:32:40.385693+00:00 : 150qpm, 4 instances
2026-07-31T03:44:55.021382+00:00 : 300qpm, 12 instances

Each test was 5 minutes, except for the first.

We should not need 4 instances to handle 90qpm (aka 1.5qps). A single api instance should be able to handle many concurrent queries each second.

This task is to investigate where the bottleneck is, and fix it.

PR #361, opened 2026-08-03
Give histograms explicit bucket boundaries

mjmor:
Part of #343 (PR G; stacked on #354). Prerequisite for the baseline threshold lines in #353 being meaningful.

A distribution's percentiles are only as precise as its buckets — a value inside a bucket is interpolated across that bucket's full width. `metrics.py` set no boundaries, so the OTel defaults applied: `…, 1000, 2500, 5000, 7500, 10000`. That leaves **four buckets above 1s**, which is the entire serving range. The repeated p95 values in the load-test tables (2425/2485, 4881/4937) are interpolation artifacts inside a single 1.5–2.5s-wide bucket, and a sub-p99 latency target above 1s cannot be evaluated at that resolution.

Ratio metrics were worse: `retrieved_share`, `kept_share` and `mean_similarity_score` are all in [0, 1], and the first default boundary is 5 — so **every value ever recorded landed in one bucket**. The same applied to `client.in_flight` / `es.client.in_flight`, where 1 through 4 concurrent requests were indistinguishable, which is precisely the low end the #344 pool-starvation signal needs.

# This PR

- Four boundary sets in `metrics.py`, selected per metric family by name suffix (`histogram_boundaries`):
  - **latency** (`*_ms`): 250 ms resolution through 3 s, 500 ms to 5 s, out to 15 s
  - **near-zero** (`eventloop.lag_ms`, `client.pool.wait_ms`, `client.connect.duration_ms`): sub-millisecond at the bottom, saturation coverage at the top
  - **concurrency** (`*.in_flight`, `*_size`): every small integer, then the pool caps
  - **ratio** (`*_share`, `*_score`): [0, 1]
- Applied via `explicit_bucket_boundaries_advisory` at instrument creation, so no Views and no per-metric wiring; unrecognised names keep the SDK default
- Each of #353's baseline thresholds (2500 ms, 100 ms, 10 ms, 100 in-flight) is an exact bucket edge, so the percentile estimate is precise at the value it is compared against — a regression test pins this, and the README says to add a boundary when changing a threshold
- Matching change in inference-service (greenearth-social/inference-service#26) so the two copies of `metrics.py` don't drift

# Testing

- `pipenv run pytest -q` — 1054 passed (24 new), on the host and **inside the devenv container**; inference-service 102 passed both ways
- Verified in devenv against a real collector export cycle: all eight sampled metrics carry their custom boundaries; 2425/2485/4881/4937 ms now resolve to ≤500 ms buckets (from 1500/2500 ms); `kept_share` 0.05/0.5/0.95 and in-flight 1/2/3 land in distinct buckets instead of collapsing into one; all four threshold values confirmed as exact edges
- `devctl feed random` renders clean, no tracebacks; both dashboard dry-runs still render OK


raindrift (approved):
This is one of those "obvious in retrospect" kind of things. Thanks for fixing it. :)

PR #380, opened 2026-08-06
Network Likes Candidate Generator in Your Feed

### LINKED ISSUE #348: Add network_likes candidate generator to main feed
Add the network_likes candidate generator to the main feed. Put it behind a posthog feature flag so we can turn it on/off easily if performance is bad. Also investigate whether there are any ways to improve its performance, since it is slow (probably; based on a small sample size, p50 = ~1s)

davidfreifeld:
Closes #348 

## This PR
Added the network_likes candidate generator to your-feed, unranked-your-feed, and cutoff-preview, behind a PostHog feature flag: "network-likes-in-your-feed". The default is for this to be enabled. This turns on the network_likes generator for local development, which in the general case doesn't connect to PostHog (I think). 

unranked-your-feed and cutoff-preview used the same candidate generators so it seemed to make sense to have them follow the behavior wrt network_likes.

Updated the config values of the social radius presets. The network_likes generator is somewhere between "friends" and "everyone", so it is most prevalent in the middle of the slider, and least prevalent on the ends.

We now call PostHog to get both feature flags asynchronously - I think this is the best way to do it vs calling PostHog twice, where each call is blocking.

I attempted to make the network_likes CG more efficient via: 
- Pass the exclude_uris into the likes query. This also prevents them from being hydrated in the next step. 
- Use `posts_recent`, since we have a one week lookback window anyway.
- Apply the one week lookback window to the likes, since if we're applying it to the posts query, we shouldn't care about any likes older than a week anyway.
- Added timed() wrappers for logging

The cold-start feed was previously using the social radius presets functionality. It would have made the code more complicated to wire through the feature flag stuff, so instead I made a change I'd been wanting to make anyway: make the cold-start feed show what a user with no likes AND no follows would see (so I removed the followed_users and two_tower_empty_history cg's). And I added a feed that just shows two_tower_empty_history candidates. Each answers its own question - the first: what does a real brand-new user see? and the second: what does the two tower model return for a user with no like history? Neither is dependent on the social radius control.

## Testing
- Added unit tests
- Ran locally
- Deployed to stage
  - Confirmed feature flag turns network_likes on and off

### DIFF
--- a/src/app/lib/candidates/network_likes.py
+++ b/src/app/lib/candidates/network_likes.py
@@ -39,6 +39,9 @@ LIKED_POSTS_PAGE_SIZE = 100
 # Hard cap on how many like documents to scan while looking for post hits.
 MAX_LIKES_SCANNED = 5_000
 
+# Lookback window for finding the matching posts
+MAX_AGE_HOURS = 168
+
 
 # ---------------------------------------------------------------------------
 # Query helper
@@ -56,6 +59,7 @@ async def fetch_recent_liked_post_uri_page(
     user_dids: list[str],
     size: int,
     search_after: list[Any] | None = None,
+    max_age_hours: MaxAgeHours = MAX_AGE_HOURS,
 ) -> LikedPostUriPage:
     """Return one page of recently liked post URIs for the given users."""
     if not user_dids or size <= 0:
@@ -63,7 +67,10 @@ async def fetch_recent_liked_post_uri_page(
 
     query = {
         "bool": {
-            "filter": [{"terms": {"author_did": user_dids}}],
+            "filter": [
+                {"terms": {"author_did": user_dids}},
+                {"range": {"created_at": {"gte": f"now-{max_age_hours}h"}}},
+            ],
         }
     }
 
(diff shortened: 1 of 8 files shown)


src/app/lib/candidates/network_likes.py
```diff
@@ -130,7 +141,7 @@ async def fetch_posts_by_uris(
     }
 
     resp = await es.search(
-        index="posts",
```
mjmor, line 133: I expect this to be a pretty big win.

mjmor (approved):
lgtm! Added @Gautham-Raju to double check the default configs for feeds, but it looks like what we discussed in chat, so don't think you should block on it.

Gautham-Raju (approved):

mjmor:
btw @davidfreifeld here are a couple perf dashboards if you want to monitor performance of our feeds during rollout:
- Deep view of everything from API server telemetry down to ES: [Per & Bottleneck Attribution Dash](https://console.cloud.google.com/monitoring/dashboards/builder/bc3f9e44-f491-455c-a910-79b9a27f2618;duration=P2D?project=greenearth-471522)
- High level view of just API server telemetry (feed and CGs): [API & Feeds](https://console.cloud.google.com/monitoring/dashboards/builder/4bb98ead-1c1f-4700-b0a0-b7266237453f;duration=P2D?project=greenearth-471522)

PR #383, opened 2026-08-07
Clear seen-posts bucket on preferences save

### LINKED ISSUE #382: "Refresh Your Feed To See Changes" should re-generate the last feed
In the UI controls pane, when you change a control you get a notification "Refresh Your Feed To See Changes"

- clicking/tapping on this should switch you back to bluesky (in the browser on desktop, back to the app on mobile)
- We should also delete the most recent batch of "seen this post" records when pressed, so that as the user adjusts controls and switches between UI and feed, they can see the effects on the *same* set of candidate posts

inthree3:
## Summary

- Adds `delete_most_recent_seen_bucket()` to `lib/firestore.py`. It deletes the current UTC day's seen-posts Firestore bucket for a user. **Idempotent**: no-op if the bucket doesn't exist.
- Calls it from `PUT /api/feeds/preferences` (`feed_transparency.py`) so that saving feed control changes automatically clears the day's seen-posts.

## Test plan

- [ ] change a feed control in the UI, verify `seen_posts/{today}` document is absent in Firestore after the preferences PUT, then reload the feed and confirm fresh candidates appear

Closes #382


raindrift (approved):
👍

PR #386, opened 2026-08-07
Cache the popularity candidate pool in Firestore

### LINKED ISSUE #330: Cache results from the Popularity CG
We use the popularity CG for a lot of cold-start users. It also imposes quite a lot of ES query load (takes ~1.5 seconds to return results). Once we fix two-tower, it'll be the slowest CG by far.

But, the result is the same for everyone. We should cache it someplace. We can have requests invalidate that cache (at the cost our p99), or we could even have a background process take the hit. Posts take some time to get popular, so it would be ok to hold on to this cache for 5 minutes.

Where should the cache go? It can contain complete candidates with all their metadata as a single document. Redis is the obvious place for a cache like this, so this task could drive out setting up some managed redis.

If we want to avoid that dev/admin overhead for now, we could also write it as a single firebase document (assuming the entire serialized candidate set is smaller than the firebase doc size limit, which I think is a megabyte).

Choose a serialization format for it that is fast and secure. json is fine, unless we can do better with something else. It doesn't need to be human-readable.

The cached object should include the time it was generated, and consumers of it should log the freshness when they use it. That way we know if it is ever getting stale. We should probably alert if it's ever > 20 mins old.

raindrift:
Closes #330.

The popularity generator's Elasticsearch query is the most expensive one on the feed path (~1.5s) and runs for a large share of requests, but its result is identical for every user — the only per-user input is the exclusion list, which is a filter over the result rather than an input to the ranking. So this keeps one pool of candidates shared by every user and every instance, and filters it per request.

Three layers, cheapest first: an in-process copy (30s), the shared Firestore entry (5 min), and Elasticsearch only on a refresh.

### Refreshing

Online rather than in a separate job, per the direction on the issue. A request that finds a stale entry is served that entry immediately and spawns a detached refresh, so no user request waits on the query. Only one refresh runs at a time: an in-process key set within an instance, and a `refresh_started_at` lease stamped inside a Firestore transaction across instances, with an expiry so a process killed mid-refresh doesn't freeze the entry.

Nothing about the uncached path gets worse. With no entry at all (cold start, or a freshness window nobody has asked for yet) the caller queries Elasticsearch for just the candidates it needs, exactly as before, and one of those callers populates the pool in the background.

### Shape of the cached data

One document per (freshness window, `video_only`) combination, created on demand — both inputs change the query, since `max_age_hours` sets the hard window *and* scales the recency decay. The pool is stored as one gzipped-JSON blob (~55 KB for 500 candidates locally; JSON so it stays inspectable and needs no deserialization allowlist). The format string doubles as a schema version, and an oversized pool has its tail trimmed rather than being rejected at write time.

Exclusions move from the query's `must_not` into an in-memory filter over the pool — that is what lets one pool serve everyone. If a user's exclusions eat through a *full* pool, generation falls back to a direct query rather than short-change the slate; a pool that came back short of its target is served as-is, because Elasticsearch has nothing more to give.

### Two things to know before merging

**Deploy `greenearth-social/frontend#71` first.** `payload` needs a Firestore single-field index exemption — Firestore indexes every field by default and rejects writes whose index entries exceed 7.5 KiB. The emulator does not enforce that limit, so a missing exemption passes every local and CI check and fails only in stage/prod.

**The alert from the issue still needs creating by hand.** There are no alert policies as code in this repo, so this emits the signal rather than the policy: `candidates.popularity_cache.age_seconds` on every lookup, `lookup_count` / `refresh_count` labelled by outcome, and an ERROR log at 20 minutes. Note that stage and prod log at WARNING, so the per-use freshness the issue asked to log lands in the metric rather than the logs.

### Verification

1184 tests pass (43 new). In the local dev environment: the first render queried Elasticsearch and stored 500 candidates; the second logged no `es_popularity` line at all, and after restarting the api container a fresh instance served from the shared pool without touching Elasticsearch. The transactional lease is faked in unit tests, so it was also exercised against the real Firestore emulator — a stale entry is claimed, an immediate second claim is refused, the merge preserves the payload so stale readers keep being served, and a fresh entry is not claimed.


src/app/lib/candidates/popularity_cache.py
```diff
@@ -0,0 +1,453 @@
...
+
+POPULARITY_CACHE_COLLECTION = "popularity_cache"
+
+# Encoding of the stored blob. Bump the suffix on any incompatible change:
+# entries written in an unknown format are ignored rather than mis-parsed.
+PAYLOAD_FORMAT = "gzip-json-v1"
+
+# Firestore's hard document limit is 1 MiB. Stay well under it — a pool that
+# would exceed this is truncated rather than rejected at write time.
+MAX_PAYLOAD_BYTES = 800_000
```
mjmor, line 60: Do we have a sense of whether we're overfetching with the current configuration? I guess since the cache-fill is off the hot path it's not a high priority to sort out though, as long as it's fast enough.

mjmor (approved):
lgtm! just a minor question

src/app/lib/candidates/popularity_cache.py
raindrift: I think it'll fit as-is. There's some code to trim it though (which is what this constant is for), and it'll log if the candidate pool is too big so we can adjust the params if needed.

mjmor: oh yea - I meant: is the candidate pool consistently too big, in which case, we could reduce the limit in the request sent to ES and make the ES query faster / less load on ES.

We'll find from the logs tho

PR #387, opened 2026-08-07
Admin API endpoints should require additional permissions

### LINKED ISSUE #367: Admin API endpoints should require additional permissions
Right now our backend has API keys. The keys don't have specific permissions associated with them though. It's all flat. If you have a key, you can access any authenticated API.

The trouble is that now we have two levels of authenticated API. The ones for external developers (like /rank/models) and the ones for our team (like /admin/*). But afaict, you can access them all once you have an api key. This will, of course, become more of a security issue if we create more API keys, and especially if we start letting people make their own.

For now we don't need a lot of resolution in our permissions, and I think we can solve this by adding an `admin` flag to the api keys. Then we can add a RequireAdminApiKey to security.py, and use it on /admin/*. Only people on the greenearth team should get admin keys, of course.

If we had a lot of users with api keys, this would be very high-priority. As things stand, it's a security issue that we should be sure to not lose track of, and fix in pretty short order (like within the next few days).

mjmor:
Closes #367

Right now any valid API key can hit /admin/* endpoints, same as external-developer endpoints. This adds a per-key admin flag so only greenearth team keys can access admin routes.

# This PR

- Add `is_admin: bool` (default `False`) to `ApiKeyDocument` — existing keys without the field parse as non-admin
- Add `verify_admin_api_key` / `RequireAdminApiKey` to `security.py`, returning 403 for authenticated-but-non-admin keys (401 still applies for missing/invalid keys)
- Switch `/admin/redirects` (POST/PUT/DELETE) to `RequireAdminApiKey`
- Add `--admin` flag to `scripts/apikeys.py generate`; `list` now shows the admin column
- Update README docs for the admin flag and CLI usage

# Testing

- `pipenv run pyright` and `pipenv run pytest -v` — 1159 passed
- Verified inside the real devenv stack (`devctl test api` — all passed)
- Exercised the real API + Firestore emulator end-to-end via `scripts/apikeys.py` and `curl`:
  - non-admin key on `POST /admin/redirects` → 403
  - admin key on `POST /admin/redirects` → 201
  - no key on `POST /admin/redirects` → 401
  - non-admin key on a regular endpoint (`GET /`) → 200 (unaffected)

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,10 @@ gea_<8-char key_id><48-char secret>
 The plaintext key is shown **once** at generation and never stored.
 Authenticate requests by passing the key in the `X-API-Key` header.
 
+Keys also carry an `admin` flag, required to access `/admin/*` endpoints.
+Existing keys default to non-admin. Only issue admin keys to the greenearth
+team.
+
 ### CLI commands
 
 Run all commands from the `api/` directory:
@@ -133,6 +137,9 @@ Run all commands from the `api/` directory:
 ```bash
 # Issue a new key
 pipenv run python scripts/apikeys.py generate alice@example.com
+
+# Issue an admin key (greenearth team only)
+pipenv run python scripts/apikeys.py generate alice@greenearth.social --admin
 
 # List all keys
 pipenv run python scripts/apikeys.py list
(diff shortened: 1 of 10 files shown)


mjmor:
FYI @inthree3 - this will break the existing API key you're using for CRUD. You'll have to mint an admin key after this lands in stage/prod.

raindrift (approved):
Looks good! Thank you!

raindrift:
Also FWIW, it should be straightforward to add the is_admin flag to whatever api key you're using. You can do it in the firebase console. Might be easier than reissuing it.

PR #397, opened 2026-08-11
Cache followed users per user in Firestore

### LINKED ISSUE #83: Cache followed users
The followed users candidate generator naively calls the bsky api to retrieve the users that the requesting user follows each time it makes a request. We should cache them to speed things up. See PR #73 for further discussion on this

If we involve the ingest pipeline, we can build a relatively robust cache. We can fetch and cache all of a user's follows on their first request, and then we can keep the cache up-to-date using jetstream ingest, since it should include follow and unfollow events.

We should probably refresh the cache periodically anyway though. This could happen with a background job, or on-demand after the cache has reached some age (a week?) This way we eventually recover if, for example, we miss a record on jetstream.

This is a story to:
 
- Implement the caching layer, in firebase or elasticsearch, whichever seems faster. Since it's a cache we don't care about durability
- Keep the cache up-to-date using the jetstream follower
- Expire records fro the cache periodically
- Use the cache when fetching followed users

mjmor:
Closes #83

The `followed_users` and `network_likes` candidate generators each walked `app.bsky.graph.getFollows` live on every feed request — up to ten sequential round-trips to `public.api.bsky.app` under a 1s budget, in front of the Elasticsearch query. When the budget expired the walk was silently truncated; when Bluesky hiccuped the generator returned `[]` and lost its whole allocation.

Follows change rarely, so they are now cached in Firestore (not Elasticsearch — ES is the loaded resource the caching is meant to protect). One document per user holding a plain list of DIDs.

# This PR

- Adds `lib/followed_users_cache.py`: read the entry, serve it, and refresh in the background under a transactional lease. No user request ever waits on Bluesky.
- Adds `FollowsFetch(dids, complete)` to `lib/bsky.py`. `complete` is true only when the cursor was exhausted or the 1000 cap was reached — not when a page failed or the budget expired. `get_followed_user_dids` still returns a plain list, so its existing callers and tests are unchanged.
- Persists that flag on the document. An incomplete entry is **always** stale however young, is still served, and refreshes on every read until one walk finishes. A partial refresh may never overwrite a complete entry, and a failed refresh clears the lease immediately (with a 60s cooldown) so the next request can retry at once.
- Migrates both call sites; consolidates the two duplicated `MAX_FOLLOWED_USERS` constants. Within one request the two generators share a lookup via the existing `RequestCache`.
- Merges `pending_adds` (written by jetstream, greenearth-social/ingex#455) at read time, so a new follow shows up before any refresh runs.
- Degrades to the pre-cache path everywhere: Firestore down means a live walk, Bluesky down means the last known follow list.

No in-process layer and no compression, deliberately — unlike `popularity_cache`, one small document per user is read about once per that user's request, so a local copy buys little and adds a second staleness window that can mask a jetstream delta.

Pairs with greenearth-social/frontend#73 (index exemption + TTL policy) and greenearth-social/ingex#455 (jetstream deltas). Neither blocks this one; deploy order is frontend → api → ingex.

# Testing

- `pipenv run pyright` clean; `pipenv run pytest -v` 1274 passed (30 new).
- devenv, `followed-users` feed: first render logged a 984ms live walk and stored 498 follows with `complete=True`; second render logged **3.7ms** and no Bluesky traffic.
- Inspected the emulator document: plain readable DIDs, `complete=true`, `expires_at` 30 days out.
- Cross-repo loop against the real emulator: the ingex Go writer appended a DID via `ArrayUnion`, and this cache served 499 follows with the new DID merged in; a `invalidated_at` stamp was then served stale (499) while the background refresh re-walked Bluesky, wrote 498, and cleared `invalidated_at`, `pending_adds` and the lease.
- No tracebacks in the api container logs.

# Telemetry (follow-up commit)

The first cut measured availability but not correctness. Added:

- **`follows_cache.age_seconds` now records on every served entry**, not only stale ones. As it stood the series contained nothing but overdue entries, so the freshness of a healthy population was invisible and the average read as alarming even when nothing was wrong.
- **`follows_cache.refresh_added` / `refresh_removed`** — how far each refresh moved the set from what was being served. This is the only correctness signal in the system: an entry that is confidently `complete` but *wrong*, because jetstream deltas were dropped and the TTL hadn't fired, is otherwise indistinguishable from a correct one. Steady state is near-zero churn; a rising delta at TTL boundaries means the live update path isn't working.
- **`follows_cache.users_rate{state=total|incomplete|stale}`** — a per-user population gauge. The lookup counters are request-weighted, so one heavy user stuck incomplete looks identical to many users briefly incomplete; this is the view that answers "is follow state broadly correct or degraded?". It runs as a **deferred task after the response is returned** — three Firestore count aggregations, never on the request path — rate-limited to one pass per `GE_FOLLOWS_CACHE_SWEEP_SEC` (default 300) per process, and a sweep failure can never affect a lookup.

Suggested alert shape: `users_rate{state="incomplete"} / users_rate{state="total"}` sustained above a few percent, and `refresh_count{outcome=~"failed|error|partial_discarded"}` non-zero over a window.

Paired with the writer counters in greenearth-social/ingex#456 — without those, a Firestore rejecting writes would leave this side looking perfectly healthy while its entries silently went stale.

Verified: 1315 tests pass (7 new). The population sweep was exercised against the real Firestore emulator with a deliberately mixed population (total=4, incomplete=1, stale=1), since the aggregation-query result shape is exactly what a fake can get wrong; and the deferred hook was confirmed firing from a live feed render, once, then rate-limited.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -96,3 +96,8 @@ export GE_FOLLOWS_CACHE_LEASE_SEC="30"
 # Quiet period after a failed refresh, so a user whose follows cannot be
 # fetched does not spawn a refresh task on every request.
 export GE_FOLLOWS_CACHE_RETRY_COOLDOWN_SEC="60"
+
+# How often one process samples population health (how many cached users are
+# incomplete or stale) and reports it as gauges. Runs deferred, off the
+# request path.
+export GE_FOLLOWS_CACHE_SWEEP_SEC="300"
--- a/.env.example
+++ b/.env.example
@@ -77,3 +77,22 @@ export GE_SERVICE_NAME="greenearth-api"
 
 # How often (in seconds) to flush metrics to Cloud Monitoring.
 export GE_METRICS_EXPORT_INTERVAL_SEC="60"
+
+########## Followed-Users Cache ##########
+
+# Per-user cache of Bluesky follow lists (issue #83). Jetstream ingest keeps
+# entries current; these knobs only govern the backstop refresh.
+
+# Age at which an entry is refreshed in the background (default 6 hours).
+export GE_FOLLOWS_CACHE_TTL_SEC="21600"
+
+# Budget for a background follow walk. Larger than the request-path clamp
+# because nothing is waiting on it.
+export GE_FOLLOWS_REFRESH_TIMEOUT_SEC="15"
+
+# How long one instance may hold the refresh lease before another may retry.
+export GE_FOLLOWS_CACHE_LEASE_SEC="30"
+
+# Quiet period after a failed refresh, so a user whose follows cannot be
+# fetched does not spawn a refresh task on every request.
+export GE_FOLLOWS_CACHE_RETRY_COOLDOWN_SEC="60"
--- a/src/app/documents.py
+++ b/src/app/documents.py
@@ -147,6 +147,61 @@ class FeedCacheDocument(BaseModel):
     )
 
 
+class FollowedUsersCacheDocument(BaseModel):
+    """Cached Bluesky follow list for one user.
+
+    The document ID is the user's ``user_doc_id``.  Replaces a live
+    ``app.bsky.graph.getFollows`` walk on every feed request (see issue #83).
+
+    Two writers touch this document.  The API owns ``follows`` and rewrites it
+    on refresh; the jetstream ingester only ever appends to ``pending_adds``
+    (via ``ArrayUnion``, so it never reads first) or stamps ``invalidated_at``.
+    That split is what keeps the two writers from clobbering each other.
+    """
+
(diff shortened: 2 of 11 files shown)


mjmor:
FYI @raindrift @davidfreifeld : I think there's probably a more elegant way to architect this (e.g., ingest handles majority of writes, scheduled GCP job periodically syncs via separate process) but this is the fastest thing to get this out the door, and that's what I'm prioritizing for now.

Also, we're preferring availability over consistency in this PR. Note how we handle exceptions. We'll need to keep an eye on this to make sure consistency is achieved or at least not too incorrect for too long.

PR #399, opened 2026-08-11
Cross-request user features cache

### LINKED ISSUE #391: Cross-request cache for user-embedding inputs
We already have a request-level cache, but we have 4-15 API instances running at once so no gurantee a request cache benefits on next user-request. A 10min TTL should be reasonable w/o impacting UX reactivity to engagement. Be careful to cache the hydrated history (uris, embeddings, author DIDs, like counts) rather than, or in addition to, the embedding because the inputs are used in later stages of the pipeline too (and make sure the cache abstraction continues to get leveraged downstream.

davidfreifeld:
Closes #391 

## Summary

This PR adds a shared, per-user Firestore cache for the recent-like features consumed by the two-tower generator and heavy ranker.

Previously, both model paths independently fetched the same recent likes and hydrated their post embeddings and metadata from Elasticsearch. They now share one ordered `UserHistory` result within a feed request and can reuse it across requests and API instances.

## Cache behavior

The cache has a bounded stale-while-refresh lifecycle:

- **Fresh:** For the first 10 minutes by default, return the cached history directly.
- **Stale:** Between 10 and 30 minutes, return the cached history immediately and refresh it in the background.
- **Hard expired:** At 30 minutes, stop serving the entry and synchronously rebuild it from Elasticsearch.

A cold request still waits for Elasticsearch because recommendation models need the history before they can run. The resulting Firestore write happens in the background, outside the recommendation latency budget.

While that write is pending, the same API instance retains the fresh value in memory so another request does not repeat the Elasticsearch work.

## Request sharing

The two-tower generator and heavy ranker now call the same `fetch_user_history_features()` helper.

The existing request-scoped single-flight cache collapses concurrent calls from both model paths into one lookup or Elasticsearch rebuild. The cached representation preserves:

- Recent liked-post URIs and timestamps in source order
- MiniLM-L12 embeddings when available
- Author DIDs
- Like counts
- Likes whose post features could not be hydrated

## Refresh coordination

Stale refreshes use two levels of coordination:

- An instance-level task registry prevents duplicate work within one API process.
- A transactional Firestore lease prevents multiple Cloud Run instances from refreshing the same user simultaneously.

Failed refreshes release the lease and record a retry cooldown, preventing every request from immediately retrying a failing Elasticsearch operation. Stale history remains available until the configured hard serving cutoff.

Background tasks and pending values are owned by the cache instance. The application stores that instance on `app.state` and drains outstanding writes and refreshes during shutdown before closing Firestore and Elasticsearch clients.

## Failure and latency behavior

The cache is an optimization and fails open:

- Firestore reads are bounded to 500 ms so cache latency cannot consume the model execution budget.
- Read failures and timeouts fall back to Elasticsearch.
- Cold-cache persistence happens asynchronously.
- Background Firestore operations have a separate 5-second timeout.
- Stale Elasticsearch refreshes have a configurable timeout.
- Refresh and write failures do not prevent serving an available history value.

## Document safety

User-history documents include:

- Schema version
- User DID and history limit
- Embedding model key
- Fetch and expiration timestamps
- Refresh lease and failure timestamps
- API release SHA
- Ordered history items and their encoded features

Documents are rejected as cache misses when their schema, user, history limit, embedding model, timestamps, or embeddings are invalid.

Embedding decoding now validates base64 input, requires the expected 384-dimensional MiniLM-L12 vector, and rejects non-finite values.

## Configuration

The deployment script stamps these values onto the Cloud Run revision while allowing environment overrides:

| Variable | Default | Purpose |
|---|---:|---|
| `GE_USER_HISTORY_CACHE_TTL_SEC` | 600 | Fresh serving window |
| `GE_USER_HISTORY_CACHE_MAX_AGE_SEC` | 1800 | Hard serving cutoff and Firestore expiration |
| `GE_USER_HISTORY_CACHE_LEASE_SEC` | 30 | Cross-instance refresh lease |
| `GE_USER_HISTORY_REFRESH_TIMEOUT_SEC` | 10 | Background Elasticsearch refresh budget |
| `GE_USER_HISTORY_CACHE_RETRY_COOLDOWN_SEC` | 60 | Delay after a failed refresh |

The configured maximum age is clamped so it cannot be shorter than the fresh TTL.

## Deployment

The frontend Firestore configuration must be deployed before this API revision. See [this PR](https://github.com/greenearth-social/frontend/pull/74). User-history documents contain large encoded embedding arrays and require the `user_history_cache.*` indexing exemption. The frontend deployment also enables native TTL on `user_history_cache.expires_at`.

If the API is deployed first, feed generation continues through Elasticsearch, but cache writes cannot populate successfully.

## Validation

- Full API test suite: **1,362 passed**
- Pyright: **0 errors**
- Focused Ruff and formatting checks passed
- Shell syntax and `git diff --check` passed
- Ran in local devenv and stage-deployed to test caching behavior manually


.env.example
```diff
@@ -101,3 +101,22 @@ export GE_FOLLOWS_CACHE_RETRY_COOLDOWN_SEC="60"
 # incomplete or stale) and reports it as gauges. Runs deferred, off the
 # request path.
 export GE_FOLLOWS_CACHE_SWEEP_SEC="300"
+
+########## User-History Cache ##########
+
+# Age at which cached recommendation history is served stale and refreshed in
+# the background (default 10 minutes).
+export GE_USER_HISTORY_CACHE_TTL_SEC="600"
```
raindrift: nit: I'm not sure that it's necessary to put the config vars for this feature in the environment. It's probably ok to just have them at as constants in the code. I don't think we will ever need to edit the cache config without running a deployment, nor will it vary between different environments.

raindrift (approved):
This looks good, just had one small comment about env vars.

This is a per-user cache so I expect it to be bigger than the popularity cache, but it shouldn't be prohibitive. We will definitely want to move this kind of shared runtime state to redis once we have it. But it should be ok in firebase for now.

.env.example
davidfreifeld: good point - I'll make the change, thanks

PR #408, opened 2026-08-12
Collapse network_likes' serial likes/hydrate paging into one round trip

### LINKED ISSUE #390: Collapse network_likes' serial likes/hydrate paging into one round trip
network_likes averages ~5 serial ES calls per run (2.6 likes pages + 2.6 hydrates, p95 ~1s each) inside a 4s budget — worst CG failure rate at 12.1%. The paging exists for likes/posts index skew, not bsky limits, so overfetch in one round trip instead.

Run a single query instead of paginating

mjmor:
Closes #390

network_likes averaged ~5 serial ES calls per run (2.6 likes pages + 2.6 hydrates, p95 ~1s each) inside a 4s budget. The paging existed to absorb likes/posts index skew, not a bsky limit, so this overfetches once instead of discovering how far to page.

# This PR

Replaces the `search_after` paging loop with a single likes scan followed by a single hydrate — 2 ES round trips regardless of index skew — and instruments the under-fill risk that trade introduces. Specifically:

- Replace `fetch_recent_liked_post_uri_page` (and `LikedPostUriPage`) with `fetch_recent_liked_post_uris`, which takes no cursor and returns URIs directly
- Scan `liked_post_scan_size(num_candidates)` likes in one query: `num_candidates * 20`, floored at 1000 and capped at the existing `MAX_LIKES_SCANNED` (5000). The old loop averaged ~260 likes to fill 30 candidates (2.6 pages x 100), i.e. ~1 surviving candidate per 9 likes; the factor of 20 is ~2.2x that yield and the 1000 floor is ~3.8x the average scan, so the tail that was driving the extra pages is covered in one shot. Enlarging this query is cheap: `_source` is one keyword field, the filter is terms + range, the sort is an indexed date
- Rank all scanned URIs by like count desc, then last-seen like recency, and hydrate only the top `hydrated_uri_limit(num_candidates)` (`num_candidates * 10`, floored at 300, capped at 1000). Hydrated docs carry embeddings, so the hydrate is deliberately decoupled from the scan — otherwise this trades 3 round trips for a much fatter fetch phase. 300 URIs at `num_candidates=30` is what the old loop hydrated in total across its pages, so fetch cost stays in today's range
- Ranking semantics are unchanged (like count desc, last-seen recency tie-break), but like counts are now computed over the whole scan rather than only the pages visited before the loop exited, so scores are a stronger signal

## Telemetry

The factors above are derived from the mean in #390 (2.6 pages), not from a measured p95 page count or per-URI hit rate, so the open question is whether some users now get fewer candidates than the paging loop would have found. These metrics answer it:

- `candidates.generate.fill_share` (labels: `generator_name`, `is_infill`) — share of a generator's requested allocation actually returned. Not generator-specific by design: it makes under-fill visible for every generator and gives network_likes a baseline to be compared against. Recorded on the primary path only — the infill path deliberately requests `shortfall * 2`, so a fill share against that denominator would read as chronic under-fill by construction
- `candidates.network_likes.likes_scan_saturated_count` — the scan returned a full window, i.e. more recent likes existed than it covered (the scan size is the binding constraint)
- `candidates.network_likes.hydrate_truncated_count` — the scan found more unique URIs than one hydrate query carries (the hydrate limit is the binding constraint)
- `candidates.network_likes.hydrate_hit_share` — measured likes-to-candidates yield, i.e. the 1-in-9 assumption both limits are sized against

Together these separate "this user has thin network activity" from "our limits cut them off", and say which limit to move.

# Testing

- Full suite on the host: `pytest -q` -> 1353 passed; `pyright` on the changed files -> 0 errors; `ruff check` on the changed files adds no new findings (the 7 reported are pre-existing lines)
- Full suite inside the devenv api container (instance `issue390`, own worktree, shared read-only ES): `devctl --name issue390 exec api pipenv run pytest -q` -> 1348 passed (run against the first commit; the telemetry commit adds no ES surface)
- Real ES client / real queries inside the container, 134 followed DIDs sampled from the likes index, `num_candidates=30`:
  - `max_age_hours=168` -> 1 ES call (`likes` size=1000), 0 candidates, no hydrate issued
  - `max_age_hours=720` -> 2 ES calls (`likes` size=1000, `posts_recent` 300 terms), 49ms, 30 candidates, like-count scores up to 4.0
- Real HTTP path: `POST /candidates/generate` with `network_likes` -> 200, `es_network_likes_search` timing line clean, no traceback in `devctl logs api`
- New metric names recorded through a real `MetricCollector` (unit tests use a fake): all four export, and both `_share` metrics pick up `RATIO_BOUNDARIES` rather than the SDK default that buckets all of [0, 1] together
- `devctl feed network-likes` returns empty on this instance because the shared fixture's newest post is 309h old and the API caps `max_age_hours` at 168 — pre-existing data drift, reproduced identically before the change; not re-seeded, since the ES data is shared with other instances
- Baseline comparison on the same data: `main` made 2 calls (`likes` 100, `posts_recent` 92) and topped out at score 2.0. The dev fixture is dense enough that `main` fills its quota on page one, so it does not reproduce the ~5-call production case; the new path is 2 round trips in both the dense and the skewed case

### DIFF
--- a/src/app/lib/candidates/generate.py
+++ b/src/app/lib/candidates/generate.py
@@ -150,6 +150,16 @@ async def run_generate(
                     generator_name=spec.name,
                     is_infill="false",
                 )
+                # Share of the requested allocation actually returned, so a
+                # generator that quietly under-fills for some users is visible
+                # without reading the feed slate.
+                returned = sum(len(result.candidates) for result in results)
+                mc.record(
+                    "candidates.generate.fill_share",
+                    min(returned / count, 1.0),
+                    generator_name=spec.name,
+                    is_infill="false",
+                )
             return [
                 result.model_copy(
                     update={"status": "empty", "reason": result.reason or "no_candidates"}
--- a/src/app/lib/candidates/generate_test.py
+++ b/src/app/lib/candidates/generate_test.py
@@ -311,6 +311,21 @@ class TestGeneratorTimeout:
         _, _, attrs = success_calls[0]
         assert attrs == {"generator_name": "popular", "is_infill": "false"}
 
+    @pytest.mark.asyncio
+    async def test_success_records_fill_share_metric(self, monkeypatch):
+        """A generator that under-fills its allocation is visible per generator."""
+        _stub_generators(monkeypatch, {"popular": _EmptyGenerator("popular")})
+        mc = FakeMetricCollector()
+        set_metric_collector(cast(MetricCollector, mc))
+
+        await run_generate(_make_request("popular", num_candidates=4), es=None)
+
+        fill_calls = [c for c in mc.calls if c[0] == "candidates.generate.fill_share"]
+        assert len(fill_calls) == 1
+        _, value, attrs = fill_calls[0]
+        assert value == 0.0
+        assert attrs == {"generator_name": "popular", "is_infill": "false"}
+
 
 # ---------------------------------------------------------------------------
 # Infill generator timeout/error tests
--- a/src/app/lib/candidates/network_likes.py
+++ b/src/app/lib/candidates/network_likes.py
@@ -21,6 +21,7 @@ from ..bsky import FollowedUsersLookupError
 from ..config import fail_fast
 from ..elasticsearch import unwrap_es_response
 from ..followed_users_cache import MAX_FOLLOWED_USERS, get_followed_dids_cached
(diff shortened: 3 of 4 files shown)


davidfreifeld (approved):
LGTM!

PR #419, opened 2026-08-14
Embeddings Hydration Exception Handling

### LINKED ISSUE #392: Embedding hydration times out under load, silently degrading ranking. Fix the exception handler.
121 "Embedding hydration failed" errors during the load test — the top ERROR in prod logs and absent from the load-test writeup. The 1.5s refetch blows its budget, so MMR falls back to author-only similarity and two-tower re-fetches. Raise or restructure.

davidfreifeld:
Closes #392 

## Summary

Embedding hydration timeouts are an expected, handled degradation, but were logged as errors with stack traces. This caused them to dominate production ERROR logs during load testing.

This change:

- Logs hydration timeouts as warnings and includes the configured timeout.
- Continues logging unexpected hydration failures as errors with stack traces.
- Preserves the existing fallback and degradation-recording behavior.
- Does not change the 1.5-second timeout or feed pipeline behavior.

Tests cover both expected timeouts and unexpected exceptions, including log severity, traceback behavior, fallback candidates, and degradation records.

## Testing

- `pipenv run ruff check src/app/routers/xrpc_test.py`
- `pipenv run pytest -q src/app/routers/xrpc_test.py`
- 214 tests passed.

### DIFF
--- a/src/app/routers/xrpc.py
+++ b/src/app/routers/xrpc.py
@@ -441,10 +441,14 @@ async def _hydrate_embeddings(es, candidates: list[CandidatePost]) -> list[Candi
                 timeout=_EMBED_HYDRATION_TIMEOUT_SEC,
             )
     except Exception as exc:
-        # If the refetch fails, MMR falls back to author-only similarity
-        # and the two-tower ranker has its own refetch path. Don't fail
-        # the request over a hydration hiccup.
-        logger.exception("Embedding hydration failed; continuing without")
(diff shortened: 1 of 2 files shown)


mjmor (approved):
I was under the impression we weren't recording the DegradationEvent for the failed embeddings correctly, but clearly we are. This is still a correct change so let's do it and close out the other issue. Thank you!

